PR #10, opened 2020-07-30
Directory page fields

stephen-cox:


modules/localgov_directories_page/config/install/field.field.node.localgov_directories_page.localgov_directory_facets_select.yml
```diff
@@ -4,8 +4,6 @@ dependencies:
   config:
     - field.storage.node.localgov_directory_facets_select
     - node.type.localgov_directories_page
-  module:
-    - localgov_directories
```
ekes, line 8: I'm guessing we want this?

Should become enforced?
```
dependencies:
  enforced:
    module:
      - localgov_directories
```

stephen-cox: I just copied the config as Drupal exported it. Yup, looks like enforcing the module dependency would be good (having just read about it https://www.drupal.org/node/2404447)

PR #25, opened 2020-09-02
Add .gitub/workflows/test.yml to trigger testing run.

finnlewis:


.github/workflows/test.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Clone drupal_container
+        uses: actions/checkout@v2
+        with:
+          repository: localgovdrupal/drupal-container
+          ref: master
+
+      - name: Create LocalGov Drupal project
+        run: composer create-project --stability dev localgovdrupal/localgov-project html
+
+      - name: Checkout ${{ github.event.repository.name }} branch to test
```
Adnan-cds, line 24: Looks like ${{ github.event.repository.name }} here is referring to the Git repository name rather than the Git branch name.

finnlewis: Hi @Adnan-cds 

As I understand it, that line just prints out what is about to happen, checking out the repo that we are testing.

The next line 

```uses: actions/checkout@v2```

checks out the commit that triggers the action, which I think in this case is the merge commit against which to run the tests.

Does that make sense? Have I understood this right?

finnlewis: https://github.com/actions/checkout/blob/main/README.md suggests that the ```ref: ''``` option defaults to the SHA of the current event if not specified.

finnlewis:
The automated checks are failing because of the missing dependency on search_api, which is a problem we are aware of in general: that new dependencies in pull requests are not noticed by composer.

.github/workflows/test.yml
Adnan-cds: Hi Finn,
If the ${{ github.event.repository.name }} mentioned in line 24 can print the Git **branch name** then all is fine.  I am saying this because rest of line 24 says "...branch to test".

Adnan-cds (approved):
:)

Issue #28, opened 2020-09-08
Tests fail which break tests for all repos


stephen-cox:

finnlewis:
I've been looking at the schema test failures, alongside @stephen-cox 

It looks like: 

 - The schema that is failing is that of the exported view's attachment view that provides the leaflet map
 - The schema appears to be defined here https://git.drupalcode.org/project/leaflet/-/blob/2.0.x/modules/leaflet_views/config/schema/leaflet_views.style.schema.yml
 - Installing config_inspector and going to https://localgov.lndo.site/admin/reports/config-inspector/views.view.localgov_directory_channel/list gives us a list of the missing schema: 

```

display.node_embed_map_attachment.display_options.style.options.data_source | Undefined | undefined | location | missing schema
-- | -- | -- | -- | --
  | display.node_embed_map_attachment.display_options.style.options.entity_source | Undefined | undefined | __base_table | missing schema
  | display.node_embed_map_attachment.display_options.style.options.entity_source_submit | Undefined | undefined | Update Entity Source | missing schema
  | display.node_embed_map_attachment.display_options.style.options.name_field | Undefined | undefined | title | missing schema
  | display.node_embed_map_attachment.display_options.style.options.description_field | Undefined | undefined | search_api_rendered_item | missing schema
  | display.node_embed_map_attachment.display_options.style.options.view_mode | Undefined | undefined | full | missing schema
  | display.node_embed_map_attachment.display_options.style.options.leaflet_map | Undefined | undefined | OSM Mapnik | missing schema
  | display.node_embed_map_attachment.display_options.style.options.height | Undefined | undefined | 400 | missing schema
  | display.node_embed_map_attachment.display_options.style.options.height_unit | Undefined | undefined | px | missing schema
  | display.node_embed_map_attachment.display_options.style.options.hide_empty_map | Undefined | undefined | 1 | missing schema
  | display.node_embed_map_attachment.display_options.style.options.disable_wheel | Undefined | undefined | 0 | missing schema
  | display.node_embed_map_attachment.display_options.style.options.fullscreen_control | Undefined | undefined | 1 | missing schema
  | display.node_embed_map_attachment.display_options.style.options.gesture_handling | Undefined | undefined | 0 | missing schema
  | display.node_embed_map_attachment.display_options.style.options.reset_map | Undefined | undefined | <array> | missing schema
[...]

PR #32, opened 2020-09-16
Search and facet block for directory

Adnan-cds:
Provides a Directory search block.  It reuses the search block from the extra field.


Adnan-cds:
Ready for review again.  To avoid the issue described in #34, do ```drush pm:install --yes localgov_directories_page``` instead of ```drush pm:install --yes localgov_directories``` while installing the localgov_directories module in a fresh localgov site.

Please note that as a result of the [fix](https://github.com/localgovdrupal/localgov_directories/pull/31/files#diff-06e5d87c1edbcfa82de0f0a23d5a6645) for #28, Directory channels now list the full Directory entry content instead of the teasers.  This is not related to this pull request.

@stephen-cox @andybroomfield @finnlewis

src/Plugin/Block/ChannelSearchBlock.php
```diff
@@ -0,0 +1,232 @@
...
+   * @var \Drupal\Core\Entity\EntityRepositoryInterface
+   */
+  protected $entityRepository;
+
+  /**
+   * Form builder service.
+   *
+   * @var \Drupal\Core\Form\FormBuilderInterface
+   */
+  protected $formBuilder;
```
stephen-cox, line 230: Very minor issue, but most classes are ordered as:
1. Class static variables
2. Class variables
4. Static methods
5. Constructor
6. Class methods

Not too bothered if you disagree :wink:

stephen-cox (approved):
The search form is wider than the sidebar so need to add a `width: 100%` to the `input` element. This could be done with something like:

```css
.block-localgov-directories-channel-search-block input[type=text] {
  width: 100%;
}
```

A couple of comments in the code, but otherwise looks fine.

css/search-block.css
```diff
@@ -0,0 +1,11 @@
...
+ * @file
+ * Style rules for directory search.
+ */
+
+/* Hide the Directory search exposed form when rendered outside our block. */
+.view-localgov-directory-channel.view-display-id-node_embed > .view-filters,
+:not(.block-localgov-directories-channel-search-block) > .views-exposed-form.localgov-search-channel-primary,
+:not(.block-localgov-directories-channel-search-block) > .views-exposed-form.localgov-search-channel-secondary {
+  display: none;
+}
```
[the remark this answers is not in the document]
Adnan-cds: The "page" display of View allows splitting off the exposed filters into a separate block.  Perhaps you are thinking of that.  In this case we are using an *Embed* display of a View which made things slightly complicated :(

Adnan-cds:
> The search form is wider than the sidebar so need to add a width: 100% to the input element.

Done :)

FAO @stephen-cox

src/Plugin/Block/ChannelSearchBlock.php
Adnan-cds: Hi Steve,
You are right that most classes follow this format.  In the absence of an agreed up on convention, I place the public methods first and the class variables last.  IIRC I picked up this convention from Pete Goodliffe's [Code craft](https://www.goodreads.com/book/show/44875.Code_Craft).  For the ChannelSearchBlock plugin class, the most important public method is ChannelSearchBlock::build().  Hence it's at the top.  The idea is, if you have to work on that block, you are most likely interested in its build() method.  So you better find it as early as possible.  I am happy to change the order.  Just let me know.

FAO @stephen-cox

stephen-cox: @Adnan-cds   Drupal core has all variables at the top followed by constructor, although static methods seem to be mixid in. Here's an example: https://git.drupalcode.org/project/drupal/-/blob/9.0.x/core/lib/Drupal/Core/Entity/ContentEntityBase.php

It's not a major thing and it sounds like this is holding other work up so happy to approve.

Adnan-cds: Thanks Steve :)

Issue #34, opened 2020-09-17
Facet related WSOD on new install of localgov_directories module


Adnan-cds:
The localgov_directories_facets facet is having trouble immediately after installing the localgov_directories module.

Error message from WSOD:
```
Drupal\search_api\SearchApiException: Could not retrieve data definition for field 'Facets' on index 'Directories'. in Drupal\search_api\Item\Field->getDataDefinition() (line 441 of modules/contrib/search_api/src/Item/Field.php).

Drupal\facets\Plugin\facets\facet_source\SearchApiDisplay->getDataDefinition('localgov_directory_facets_filter') (Line: 589)
Drupal\facets\Entity\Facet->getDataDefinition() (Line: 99)
Drupal\facets\Plugin\facets\processor\BooleanItemProcessor->supportsFacet(Object) (Line: 219)
Drupal\facets\Form\FacetForm->form(Array, Object) (Line: 149)
Drupal\Core\Entity\EntityForm->buildForm(Array, Object)
call_user_func_array(Array, Array) (Line: 532)
Drupal\Core\Form\FormBuilder->retrieveForm('facets_facet_edit_form', Object) (Line: 278)
Drupal\Core\Form\FormBuilder->buildForm(Object, Object) (Line: 91)
Drupal\Core\Controller\FormController->getContentResult(Object, Object)
call_user_func_array(Array, Array) (Line: 123)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->Drupal\Core\EventSubscriber\{closure}() (Line: 573)
Drupal\Core\Render\Renderer->executeInRenderContext(Object, Object) (Line: 124)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->wrapControllerExecutionInRenderContext(Array, Array) (Line: 97)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->Drupal\Core\EventSubscriber\{closure}() (Line: 151)
Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object, 1) (Line: 68)
Symfony\Component\HttpKernel\HttpKernel->handle(Object, 1, 1) (Line: 57)
Drupal\Core\StackMiddleware\Session->handle(Object, 1, 1) (Line: 47)
Drupal\Core\StackMiddleware\KernelPreHandle->handle(Object, 1, 1) (Line: 106)
[...]

PR #42, opened 2020-09-28
Teaser view: Defer indexing of the Facet selection field.

### LINKED ISSUE #38: Display of directory items on channel page should use teaser display mode
As mentioned in point 5 on https://github.com/localgovdrupal/localgov_theme/issues/85, the display of directory items on a channel page should use the node teaser view instead of the default view.

There is some code that sets this, but this was commented out because it breaks the tests: https://github.com/localgovdrupal/localgov_directories/blob/27112a3a05689588a6ed7f641a0332ef52d948ec/localgov_directories.module#L154-L170

The reason it's done in code rather than in the view configuration is because we don't know which directory item node types are installed when enabling the module.

Adnan-cds:
Bringing back the Views config update code.  For this to work, we have to defer
setting up indexing on the Facet selection field.  A Directory entry content
type may not have the Facet selection field when the Channel selection field is
added to it.  This breaks the Search index.  Deferring indexing of the Facet
selection field until this field is actually added to a Directory entry content
type resolves this.

Closes #38 #34


stephen-cox (approved):
Great work @Adnan-cds. Tests all pass and the correct display mode is now set during a site install. Looks like it was a tricky one to fix.

Adnan-cds:
Thanks Steve.  Yeah, there were two issues to resolve which made it tricky.  I fixed one (search index referring to non-existent facet field) and thought all is good.  And then the second one (Facet config referring to non-existent facet field) surfaced.

In this ticket, we have dealt with the non-existent Facet field.  On a **fresh install** of localgov_directories, the Search Api module is still complaining about the non-existent Directory channel selection field (i.e. field.field.node.BUNDLE-NAME-GOES-HERE.localgov_directory_channels) since that one doesn't exist initially either.  But this issue doesn't cause a WSOD and instead logs an error message with the Watchdog, so I will raise another ticket on that.

Looks like there is a merge conflict, so I will merge the master branch with this branch and then merge it back to master.

@stephen-cox

PR #44, opened 2020-09-30
Facet group heading: Clearly identify group heading.

Adnan-cds:
Facet header items should be wrapped in a header tag.  This change was going into the
localgov_theme in the form of an *overridden* Twig template.  But it looks so
appropriate that we thought it better to include it in the source template.

Issue: https://github.com/localgovdrupal/localgov_theme/issues/85

FAO @finnlewis @cjstevens78

### DIFF
--- a/templates/facets-item-list--links--localgov-directories-facets.html.twig
+++ b/templates/facets-item-list--links--localgov-directories-facets.html.twig
@@ -44,7 +44,7 @@
   {% set top_list_type_attr = create_attribute(attributes.toArray()).removeClass('js-facets-checkbox-links') %}
   <{{ list_type }}{{ top_list_type_attr }}>
   {%- for group in items -%}
-    <li{{ group.attributes }}>{{ group.title }}</li>
+    <li{{ group.attributes }}><h4 class="facet-group__title">{{ group.title }}</h4></li>
     <{{ list_type }}{{ attributes.addClass('facet-filter-checkboxes') }}>
     {%- for item in group.items -%}
       <li{{ item.attributes }}>{{ item.value }}</li>


stephen-cox:
Just a question - I don't really know the answer - what's accessibility of using h4 here? I wonder if `label` is better here.

cjstevens78:
> Just a question - I don't really know the answer - what's accessibility of using `h4` here? I wonder if `label` is better.

Yes I agree, ideally each facet set would be grouped into a `<fieldset>` with a `<label>` describing the `<input type="checkbox">` options. However, this is not how the facets have been built. They render as a `<ul>` list with no distinction between the first `<li>` which contains the title of the facet set and the following `<li>` tags. So for accessibility i've gone with adding an `<h4>` around the title so as to describe the options that come after it. An `<h4>` is a nice safe level of header tag unlikely to conflict with anything going on with the page it is used on.

Adnan-cds:
Hi Steve,
AFAIK ```label``` would be appropriate while labelling an input element whereas here we are labelling list items.  Example markup from *existing* code:
```
<ul class="...">
  <li>First facet title</li>
  <ul class="...">
    <li>Facet item 0</li>
    <li>Facet item 1</li>
    ...
  </ul>
  <li>Second facet title</li>
  <ul class="...">
    <li>Facet item A</li>
    <li>Facet item B</li>
    ...
  </ul> 
</ul>
```
This markup is a bit odd given that ul elements are direct children of another ul element (which is not [valid HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul)).  Ideally, the inner ul elements should be under li elements and not directly under the outer ul element.  But I have ignored that part in this pull request given we are only bringing in a small change from the localgov_theme.

FAO @stephen-cox @cjstevens78

stephen-cox (approved):
Just checking that accessibility has been thought about. Happy to approve.

Adnan-cds:
Thanks Steve :)  I am merging it.  I will log a ticket to improve the Facet markup.

Issue #18, opened 2020-08-05
Facet tests


ekes:
Spun out from #9 where the facets were made, but not fully tested.

Adnan-cds:
Possible test steps for a Functional test:
- Set the test theme to *localgov_theme*.
- Create a Directory facet type (e.g Location).
- Add a few Directory facet items (e.g. North, South, etc.) of the above type.
- Create a Directory channel node and tell it to use the above Facet type.
- Create another content type to serve as Directory entry.  Place this content type in the above Directory channel.
- Add a few nodes of the above content type.  Select one or more Facet items when creating each of these nodes.
- Try to view the Directory channel node.  The Facet items should appear as part of the page.
- Bonus step: click one of the Facet links and ensure only the right nodes are listed.

This doesn't have to a be a Functional test.  An appropriately written Kernel test should also do.

PR #48, opened 2020-11-04
Facet weight

### LINKED ISSUE #35: Sort Facet items by weight
A weight property should allow us to manually order Facet type and Facet items.  At the moment these are ordered alphabetically.

Sub tasks:
- [x] Add *weight* property to the *localgov_directories_facets* and *localgov_directories_facets_type* config entities.
- [x] Update [Drupal\localgov_directories\LocalgovDirectoriesFacetsListBuilder](https://github.com/localgovdrupal/localgov_directories/blob/master/src/LocalgovDirectoriesFacetsListBuilder.php) and [Drupal\localgov_directories\LocalgovDirectoriesFacetsTypeListBuilder](https://github.com/localgovdrupal/localgov_directories/blob/master/src/LocalgovDirectoriesFacetsTypeListBuilder.php) to use these weight values when listing their respective entities.
- [x] A Facet sort plugin that sorts based on the weight property instead of the current *Sort by display value*.
- [x] Update the Facet related config files.

Adnan-cds:
- Added a "weight" field to both the localgov_directories_facets content entity and localgov_directories_facets_type config entity.
- A facet sort plugin that respects the weight property of localgov_directoires_facets entities.
- An update hook to "activate" the above.
- Convinced our custom [facet template preprocessor](https://github.com/localgovdrupal/localgov_directories/blob/f68e7d5e0952534ff14317cf6a12690aba2b29cc/src/DirectoryExtraFieldDisplay.php#L249) to respect the weight property of the localgov_directories_facets_type entities.

### Improvement ideas
- ~A Functional test to verify that the Facets are indeed rendered based on their weight.  I will raise a separate ticket for this.~  This already [exists](https://github.com/localgovdrupal/localgov_directories/issues/18) thanks to @ekes 

Resolves #35

### DIFF
--- a/config/conditional/facets.facet.localgov_directories_facets.yml
+++ b/config/conditional/facets.facet.localgov_directories_facets.yml
@@ -33,12 +33,6 @@ hard_limit: 0
 exclude: false
 only_visible_when_facet_source_is_visible: false
 processor_configs:
-  display_value_widget_order:
-    processor_id: display_value_widget_order
-    weights:
-      sort: 40
-    settings:
-      sort: ASC
   translate_entity:
     processor_id: translate_entity
     weights:
@@ -50,6 +44,12 @@ processor_configs:
       pre_query: 50
       build: 15
     settings: {  }
+  weight_property_order:
+    processor_id: weight_property_order
+    weights:
+      sort: 50
+    settings:
+      sort: ASC
 empty_behavior:
   behavior: none
 show_title: false
--- a/config/conditional/facets.facet.localgov_directories_facets.yml
+++ b/config/conditional/facets.facet.localgov_directories_facets.yml
@@ -33,6 +33,12 @@ hard_limit: 0
 exclude: false
 only_visible_when_facet_source_is_visible: false
 processor_configs:
+  display_value_widget_order:
+    processor_id: display_value_widget_order
+    weights:
+      sort: 40
+    settings:
+      sort: ASC
   translate_entity:
     processor_id: translate_entity
     weights:
@@ -47,7 +53,7 @@ processor_configs:
   weight_property_order:
     processor_id: weight_property_order
     weights:
-      sort: 50
+      sort: -5
     settings:
       sort: ASC
 empty_behavior:
--- a/config/schema/localgov_directories.entity_type.schema.yml
+++ b/config/schema/localgov_directories.entity_type.schema.yml
@@ -8,5 +8,8 @@ localgov_directories.localgov_directories_facets_type.*:
     label:
       type: label
       label: 'Label'
+    weight:
+      type: integer
+      label: 'Weight'
     uuid:
       type: string
--- a/localgov_directories.install
+++ b/localgov_directories.install
@@ -28,7 +28,12 @@ function localgov_directories_update_8001() {
   $field_storage_def_for_weight = BaseFieldDefinition::create('integer')
     ->setLabel(t('Weight'))
     ->setDescription(t('The weight of this Directory facet in relation to other facets.'))
-    ->setDefaultValue(0);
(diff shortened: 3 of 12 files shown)


stephen-cox (changes requested):
Hi @Adnan-cds 

The code for this looks fine, but I have queries regarding the functionality.

It looks to be possible to order the facet types using a draggable interface but not the facet items, so to change these content editors will need to manually set this be editing the facet item. Is that correct? This is not a major problem as it would be quite a big overhaul of the facet item interface to implement a draggable way to set weights on these; although grouping facet items by their type would be a big improvement to the editor experience.

The main issue I find is the default behaviour when all the facet weights are set to 0. Before saving the facet type order these display on a directory channel page in a random order. Facet items are also randomly ordered. If content editors don't want to set the weight on these things I would expect them to display alphabetically. So there needs to be two sorts applied; weight and then alphabetical for equally weighted items.

If you can add this second sort then I'm happy to approve this.

Adnan-cds:
> It looks to be possible to order the facet types using a draggable interface but not the facet items

That's right.  Drupal core comes with a [Draggable list class](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Config%21Entity%21DraggableListBuilder.php/class/DraggableListBuilder/9.0.x) for config entities but not content entities.  That stopped me from implementing the dragging functionality for the Facet items.  It is still possible to achieve this through a custom list built on top of Drupal core's [tabledrag](https://api.drupal.org/api/drupal/core%21includes%21common.inc/function/drupal_attach_tabledrag/9.0.x) functionality.  That will be a UX improvement for editors but won't change the facet behaviour.  Perhaps we can implement this as part of a separate ticket since it is not blocking the migration of Facet items.

> If you can add this second sort then I'm happy to approve this.

Makes total sense.  I shouldn't have missed this.  I'll update the code and give you a shout.

Thanks for the feedback :)

finnlewis (approved):
Hi @Adnan-cds and @stephen-cox 

I've tested too and can confirm the functionality works, even if it is all a bit of a UX challenge for site builders to change the ordering. Hopefully we can improve the UX in due course. 

I imagine we will also need to extend the sorting options to for facets within a facet type to include default sorting by alphabetical, by weight and by number of results. The latter is a standard approach with faceted filters which I expect some people will want, such that large result sets can display facet options something like this: 

Colour: 
 - red (738)
 - blue (438)
 - green (37)

But that's for another PR. 

I'm happy with this as is for now.

Thanks @Adnan-cds !

finnlewis:
Oh - I tested the hook_update_n too - installed the module, created some content and facets, then switched to this branch and ran updb. All seemed to work as expected.

Adnan-cds:
> facets within a facet type to include default sorting by alphabetical

Thanks Finn.  I hope to restore the alphabetical sorting (after weight-based sorting) by today.  I will request you and Steve for another round of testing later.

Adnan-cds:
Ready for review once again. @stephen-cox @finnlewis 

Both Directory facet types and Directory facet items should now be first sorted by weight and then label.  This means both Facet types and Facet items with equal weights will be sorted by their labels.

In the Directory facet listing page (i.e. /admin/content/directories/facets), sorting now happens based on Facet type, next weight, and then label.  This allows Facet items of the same Facet type to all appear together in this list.  This is to improve editor experience as per Steve's suggestion.

stephen-cox (approved):
Looks good @Adnan-cds Sorting is working as I would expect it; weight and then alphabetical. The change to the sort on the fact items page is also an improvement. Happy to approve this.

Adnan-cds:
Thanks Steve, thanks Finn.  I am merging this master.

PR #50, opened 2020-11-10
Apply embeded view facets to map attachment.

ekes:
At present the facets embeded view of directory entries isn't filtering the map attachment at the same time as the list. This applies the facets from the embedded list to the map as well.

I'd pondered making this more general; but decided specific for this one case is good for now. When there are more maps it can decided to repeat this with them where ever they exist, or write a rule for naming patterns that makes it easy to apply the same rule to multiple views.

For now this fixes the immediate issue.


Adnan-cds (approved):
My lack of knowledge about the SearchAPI stops me from fully appreciating this fix.  But it works for me :)

finnlewis:
Thanks @Adnan-cds 

@ekes merge I reckon!

PR #51, opened 2020-11-11
Better ignore for Facet Types, now with tests.

ekes:
The previous built-in ignore for exporting facet type config was... uhum... a little brutal and so Drupal sometimes moaned about deleting the config; it also didn't take into account collections (mainly used for languages).

This update - with full tests - works better by replacing the configuration in the correct service with the live content. There's a quite full explanation on this 'blog' post I wrote https://gitlab.com/ekes/drupal-config-transform-event/-/blob/master/README.md The module there also has the full tests https://gitlab.com/ekes/drupal-config-transform-event 

Using this also makes it possible for sites to stage the facet types if they want. There is a $configuration setting that can be added. I've added a note about this is the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,5 +58,15 @@ The content type should be automatically added to the search index when you add
 the directory fields. You may want to create a `directory index` display mode
 as this will be used for the full text search indexing.
 
+### Staging 'Directory facet types'
+
+By default facet types are not exported to configuration and are treated as
+content an administrator user can create on production. If you want to have
+types in configuration as vocabularies would be on the site to create and
+export the facet types set:
+`$settings['localgov_directories_stage_site'] = TRUE;`
+and it will be exported with other configuration. Any types that exist in
+configuration will be imported.
+
 ## Block placement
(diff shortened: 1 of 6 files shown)


finnlewis:
Tests pass, which is a good start. 

I'll try to test this on a default install later. 

@Adnan-cds or @andybroomfield are either of you able to review also?

Many thanks,

Finn

Adnan-cds (commented):
Code looks alright.  I will test it next and confirm what happens.

Thanks for the code.  I was not looking forward to updating the config files for the Facet types every now and then they are updated by editors.  This should help :)

Adnan-cds:
This is what happened during my test:
- I have updated the title of a Directory Facet type entity from its **form** and then tried 'drush config:export'.  Nothing was exported.
- I then edited the **config file** for the same Directory Facet type entity.  There I changed the title of the Facet type.  'drush config:import' happily imported the label change.

Based on the code and Ekes' [write up](https://gitlab.com/ekes/drupal-config-transform-event/-/blob/master/README.md), I think this is how it is supposed to work.  Please confirm.

FAO @ekes

ekes:
Yep. That's the idea.

*IF* you want to control any facet types with config you can, and they will be controlled if they are in config.

But by default they don't export, and so they aren't in config, so they don't import, and they act just like content.

Adnan-cds:
Thanks for the confirmation Ekes.  All looks good.  I will just delete all the Facet type configs I current have in the disk :D

Adnan-cds (approved):
Code looks okay.  Works as expected.  So why not :)

Issue #52, opened 2020-11-14
Map attachment


ekes:
Follow-up to https://github.com/localgovdrupal/localgov_directories/pull/50

I've discovered at least the way we're adding the embed to the content causes views to run Attachment::buildRenderable which causes it to getCacheTags which down the line causes search_api to run preExecute which applies all its processors.

So it's no running the query, just getting the cache tags for the render array, but this runs the processors on the filters.

Except the way views works it doesn't get the filter arguments and set them on the display till the execute on the parent display.

The processors on the filter arguments are only run once.

So the attachment does get the filter arguments before the query is run, but without the processors run.

Which in cutting to the chase means: no stemming, no case insensitive etc. Which means the results are different for the embed and the attachment.

That's as much staring at that code as I want to do. My temptation here is suggest we move to not using the attachment, but having two embed displays or something using the same filter keys, and placing them on the same page.

ekes:
Believe this is fixed with the two MR above.

PR #53, opened 2020-11-24
Fix/52 map attachment embed

ekes:
To deal with the preprocess being run before an attachment gets its arguments #52 This changes the attachment to an embed. Adds it as an additional field (as the list). And applies the facets from the list to the embed (as #50 did for the attachment).

### DIFF
--- a/modules/localgov_directories_location/config/override/views.view.localgov_directory_channel.yml
+++ b/modules/localgov_directories_location/config/override/views.view.localgov_directory_channel.yml
@@ -181,36 +181,13 @@ display:
         - 'user.node_grants:view'
       tags:
         - 'config:search_api.index.localgov_directories_index_default'
-  node_embed:
+  embed_map:
     display_plugin: embed
-    id: node_embed
-    display_title: Embed
-    position: 1
-    display_options:
-      display_extenders: {  }
-      cache:
-        type: none
-      defaults:
-        cache: false
-    cache_metadata:
-      max-age: -1
-      contexts:
-        - 'languages:language_interface'
-        - url
-        - url.query_args
-        - 'user.node_grants:view'
-      tags:
-        - 'config:search_api.index.localgov_directories_index_default'
-  node_embed_map_attachment:
-    display_plugin: attachment
-    id: node_embed_map_attachment
-    display_title: Attachment
+    id: embed_map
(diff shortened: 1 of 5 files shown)


finnlewis:
@Adnan-cds this is the fix to make sure the maps work on directories. 

If you've got a chance to ap[prove this then we could include this in the next release. 

I've tested locally and it works much better.

The code looks good to me too.

Adnan-cds:
> The code looks good to me too.

Ditto.

Any chance we can have an update hook.  I am thinking of something like the following:
```
/**
 * Totally untested :)
 */
function localgov_directories_location_update_8001() {
 
  $view_with_map_embed = Yaml::decode(file_get_contents(drupal_get_path('module', 'localgov_directories_location') . '/config/override/views.view.localgov_directory_channel.yml'));

  $view = View::load('localgov_directory_channel');
  if ($view) {
    $display = $view->get('display');
    $display['embed_map'] = $view_with_map_embed['display']['embed_map'];
    $view->set('display', $display);
    $view->save();
  }
}
```

ekes:
Ah! Now. Just adding the display with the 'extra field' would suddenly make it appear. But maybe it's possible to also set the default on the field to off.
Then existing sites would remain with their attachment in config, but can move to using the (now in config) embed and enable the 'extra field' when ready.

Adnan-cds:
I am assuming what you are trying to say is that we don't need an update hook.  In that case this is good to go :)

Adnan-cds (approved):
Thanks :)

ekes:
> I am assuming what you are trying to say is that we don't need an update hook.

Not quite. That it would be go good idea, but that it shouldn't be enabled by default. So I added https://github.com/localgovdrupal/localgov_directories/pull/53/commits/9695f23e97ca6eaadef2ecc7b40900901f1e7261#diff-620186022870f975a77e6a9d0832dccb786b4e0ac7859218e50025bc02b8f564R43 to your suggestion.

Adnan-cds:
> So I added 9695f23

Sorry to misunderstood you previously.  I will test the latest change and try to get back to you by today.

Adnan-cds:
I have tried commit 9695f23.  I won't claim I understand this change entirely.  But it hasn't broken anything.  So I am happy for it to be merged :D

PR #55, opened 2020-12-01
Don't try to access missing facet in preprocess.

ekes:
Not something that's going to happen in LocalGov as is I don't think, because you have to add facets to a channel. I'll make another patch for allowing people not to :)

But, if you've edited the config and allow a directory with no (or different) facets and this is called when the facet hasn't been added at all before preprocess it'll through an fatal access method on null.

### DIFF
--- a/localgov_directories.module
+++ b/localgov_directories.module
@@ -251,7 +251,7 @@ function localgov_directories_field_config_delete(FieldConfigInterface $field) {
  * @see facets_preprocess_facets_item_list()
  */
 function localgov_directories_preprocess_facets_item_list(array &$variables) {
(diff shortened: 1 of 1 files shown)


finnlewis:
@Adnan-cds @andybroomfield 
This is now a live issue on Lambeth.
OK if we merge this?

tanc:
Tested on Lambeth dev site and fixes the fatal error on a directory which doesn't have any 'magic facets'

Adnan-cds (approved):
Thanks. Sorry for the delay in reviewing it :(

andybroomfield (approved):
These should be fine, but haven't had chance to do a full test.
Ideally it would be good to have a test that verifies the problem, and how the patch resolves it.

PR #62, opened 2020-12-19
Bug fix: corrects required issue where values lost, make required work

ekes:
When required the facet field lost values. It should have set the default to a single facet enabled, if there was only one.

Also required didn't actually stop the submission of the form with no facets selected.

Fixes and extended tests for different variants. Also adds if there are no facets available on a channel.

### DIFF
--- a/src/Plugin/EntityReferenceSelection/LocalgovDirectoriesChannelsSelection.php
+++ b/src/Plugin/EntityReferenceSelection/LocalgovDirectoriesChannelsSelection.php
@@ -2,8 +2,14 @@
 
 namespace Drupal\localgov_directories\Plugin\EntityReferenceSelection;
 
+use Drupal\Core\Entity\EntityFieldManagerInterface;
+use Drupal\Core\Entity\EntityRepositoryInterface;
+use Drupal\Core\Entity\EntityTypeBundleInfoInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Entity\Plugin\EntityReferenceSelection\DefaultSelection;
+use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Session\AccountInterface;
 
 /**
  * Plugin description.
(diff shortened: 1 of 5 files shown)


Adnan-cds:
Sorry Ekes, Github has done it again.  The build image (Ubuntu 18) is now using PHP 8 instead of PHP 7 and composer install is failing.  On the upside, Drupal 9.1+ is now [PHP 8 compatible](https://www.drupal.org/docs/system-requirements/php-requirements) which is an incentive to upgrade LocalGov to Drupal 9.1+.

In the short term, we can [pretend](https://getcomposer.org/doc/06-config.md#platform) that we are using PHP 7.x.

FAO @finnlewis @andybroomfield @stephen-cox

andybroomfield:
@Adnan-cds see https://github.com/localgovdrupal/localgov/issues/167
Do you know how to fix?

Adnan-cds:
Hi Andy, no, I wasn't aware that you have already [fixed it](https://github.com/localgovdrupal/.github/pull/4/files).  Thanks.  Yes, that's one of the correct fixes.  We now need to apply that fix to each and every test action file however annoying that may sound.  So for example if Ekes updates localgov_directories' [test action](https://github.com/localgovdrupal/localgov_directories/blob/master/.github/workflows/test.yml) file, this test should pass.

Adnan-cds:
BTW, some LocalGov tests [fail under PHP 7.4](https://github.com/localgovdrupal/localgov_project/pull/35), so it has to be the unsupported PHP 7.2 for now.  I haven't tried PHP 7.3 yet which is still supported.

ekes:
> Please ignore the above. The change from localgovdrupal/localgov#167 if implemented to localgov_directories and other localgov repos should work fine.

Added as a branch and merged in here. Let's see.

.github/workflows/test.yml
```diff
@@ -12,6 +12,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
```
Adnan-cds, line 18: Let's change this to '7.2' which is what the [Docker container that runs the tests](https://github.com/localgovdrupal/drupal-container/blob/596e0bc03241c240d1f7bf05ea5c53ba55a233ac/Dockerfile#L17) uses.

andybroomfield: Yes, I just tried that, so seeing how that goes in one of my tests.

finnlewis:
Yay tests pass! 

@Adnan-cds @andybroomfield either of you able to review ?

Adnan-cds (commented):
> either of you able to review 

Working on it.  Not the easiest to review.  Trying to understand what problem this fix addresses.  Hope to get back to you by the end of the day.

finnlewis:
Thanks @Adnan-cds

Adnan-cds:
Lots of changes.  Looks like more than one fix have been bundled together.  One about the Directory channel selection widget.  Another about the facet selection.  I am not confident I understand the fixes entirely although I could follow the tests thanks to the embedded comments.

In several cases, doc comments for methods are simply saying ```{@inheritdoc}``` even though the method (e.g. a form validator) has custom logic in it.  I wish it had actually said what that method is doing.

Tests are passing of course.  Andy, do you have anything against accepting this change?

@andybroomfield

finnlewis:
This is an important bug fix that is a current problem on the Lambeth site. 

I think we're just waiting in case @andybroomfield has any objections. 

If not, I'd like to merge this by the end of the day to be able to release to Lambeth on Monday and leave Lambeth directories in better shape.

Adnan-cds:
> I think we're just waiting in case @andybroomfield has any objections.
>
> If not, I'd like to merge this by the end of the day 

I was thinking the same :)

andybroomfield:
I don't have anything to add, and its not something I can check right now.
If the test is passing for this use case I would suggest we merge now, and flag and resulting bugs as new issues.
I'd imagine losing facets is critical to resolve.

Adnan-cds:
Thanks Andy.  I will approve it then.

Adnan-cds (approved):

PR #65, opened 2020-12-22
PHP version fix for github runners localgovdrupal/.github/pull/4

ekes:
Is this the same thing as https://github.com/localgovdrupal/.github Would it work as a subrepo or something?

Let's see if it passes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
+
       - name: Clone drupal_container
         uses: actions/checkout@v2
         with:


andybroomfield:
@ekes, I think actully it has to match the same as the docker container (sorry not sure how this is set up).
I have a test here with the version set to php 7.2 https://github.com/localgovdrupal/localgov_alert_banner/pull/103/checks?check_run_id=1595462906

ekes:
As 7.4 is the lowest currently actively supported we should be targeting that really.

finnlewis:
The tests are still failing with: 

```
Your Composer dependencies require a PHP version ">= 7.3.0". You are running 7.2.24-0ubuntu0.18.04.7.
```

So why is it still testing with 7.2 even though further up the test output it appears to have set up as PHP 7.4?

finnlewis:
Ah - I see now 

```
Run docker-compose up -d
Creating network "localgov_directories_default" with the default driver
Pulling drupal (localgovdrupal/apache-php:php7.2)...
php7.2: Pulling from localgovdrupal/apache-php
Digest: sha256:89de074b13d66b4aaf8e445b58b4dc95976c2bf947ee2ee9cdd85a6fb981f24e
Status: Downloaded newer image for localgovdrupal/apache-php:php7.2
Pulling chrome (drupalci/webdriver-chromedriver:production)...
production: Pulling from drupalci/webdriver-chromedriver
Digest: sha256:156e26701446857450e07cfd6d48b4987b60d6f3bc3698026c844a102003b25e
Status: Downloaded newer image for drupalci/webdriver-chromedriver:production
Creating drupal ... 
Creating localgov_directories_chrome_1 ... 
Creating localgov_directories_chrome_1 ... done
Creating drupal                        ... done
```

How do we change the docker image to use a php 7.4 image? Maybe @stephen-cox or @Adnan-cds one of you can point us in the right direction?

Adnan-cds:
> How do we change the docker image to use a php 7.4 image?

Hello Ekes, Finn, Steve,
Happy new year :)  To use PHP 7.4, we need to update our [Dockerfile](https://github.com/localgovdrupal/drupal-container/blob/master/Dockerfile) and then tell [Docker hub](https://hub.docker.com/r/localgovdrupal/apache-php) to rebuild the image.  Steve has access to our Docker hub account, so should be able to help with that.

But the real problem is, our [tests don't all pass](https://github.com/localgovdrupal/localgov_project/pull/35) under PHP 7.4 :(  So the first step would be to ensure we have sorted this out locally within Lando.

FAO @ekes @finnlewis @stephen-cox

.github/workflows/test.yml
```diff
@@ -12,6 +12,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
```
Adnan-cds, line 18: This is the  fix we want in an ideal setup where every piece of our infrastructure is up to date.  But given that our [Docker image is still stuck at 7.2](https://github.com/localgovdrupal/drupal-container/blob/596e0bc03241c240d1f7bf05ea5c53ba55a233ac/Dockerfile#L17), how about settling for 7.2 here as well?  That will at least let this test pass for now.

andybroomfield: I reverted mine in the alert banner to 7.2 until we have the 7.4 tests fixed.
https://github.com/localgovdrupal/localgov_alert_banner/pull/103/files

Adnan-cds: Thanks for the pointer Andy.  Tebb, [Andy's change](https://github.com/localgovdrupal/localgov_alert_banner/pull/103/files) is what you need for the [localgov_roles PR](https://github.com/localgovdrupal/localgov/pull/168/files).

FAO @tebb

stephen-cox:
We should certainly be using PHP 7.4. Happy to update the Docker image to support this. I can also give others access to the Docker Hub account so they can push new images. We should ensure that all tests are passing with PHP 7.4 before we push a new image to the `latest` tag though.

finnlewis:
Hey @stephen-cox !
Happy New Year! 
I'm close to getting all tests to run in PHP7.4, see https://github.com/localgovdrupal/localgov_directories/issues/67
Do please add me to the Docker Hub account, I'd like to learn how to update the docker images and perhaps I can document this at the same time.

finnlewis:
@Adnan-cds is this fgood to merge or do we want to add the latest test config from your recent work?

Adnan-cds (approved):
Looks alright.  Thanks :)

PR #71, opened 2021-01-24
Group add directory content links together if menu link groups is enabled

tebb:
This pr adds the optional localgov_menu_link_groups config to localgov_directories per this issue:

https://github.com/localgovdrupal/localgov/issues/119

Config for the other relevant modules and a functional javascript test for localgov_services is also ready.

### DIFF
--- a/config/optional/localgov_menu_link_group.localgov_menu_link_group.localgov_menu_link_group_directory.yml
+++ b/config/optional/localgov_menu_link_group.localgov_menu_link_group.localgov_menu_link_group_directory.yml
@@ -1,0 +1,16 @@
+langcode: en
+status: true
+dependencies:
+  enforced:
+    module:
+      - localgov_directories
+      - localgov_menu_link_group
+id: localgov_menu_link_group_directory
+group_label: Directory
+weight: 0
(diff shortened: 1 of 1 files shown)


tebb:
@stephen-cox : Testing passes here: https://github.com/localgovdrupal/localgov_directories/actions/runs/508103560

Can we ignore or fix the 'Annotation' at the bottom of that page?

"Failed to remove 'http.https://github.com/.extraheader' from the git config"

Adnan-cds (approved):
Works for me :)

Sorry, it took so long.

ekes:
I assume the optional config is being re-scanned, for other reasons, when you enable a submodule like localgov_directories_venue after having already installed other things, so this gets picked up if you do it in that order.
 
Alternative would be to put the config into the submodules - would that work - would it be better?

Adnan-cds:
> Alternative would be to put the config into the submodules - would that work - would it be better?

That should work.  But as you have said, it already works the way @tebb has arranged it.  Ideally each module should have its own copy of a Menu group config file.  Because localgov_directories_page and localgov_directories_venue are submodules, I am less bothered.

PR #74, opened 2021-04-22
Hide duplicate exposed filter form

Adnan-cds:
When the exposed filter form *block* for the Directory listing view is already
present in the page, do not display another similar exposed filter form above the **Map** display of
that view.

In other words, we are ending up with two similar Exposed filter forms - one from the Exposed filter form **block** in the sidebar, another from the Map display of the localgov_directory_channel view.  This change hides the second one.

### DIFF
--- a/css/search-block.css
+++ b/css/search-block.css
@@ -5,6 +5,7 @@
 
 /* Hide the Directory search exposed form when rendered outside our block. */
 .view-localgov-directory-channel.view-display-id-node_embed > .view-filters,
+.view-localgov-directory-channel.view-display-id-embed_map > .view-filters,
 :not(.block-localgov-directories-channel-search-block) > .views-exposed-form.localgov-search-channel-primary,
 :not(.block-localgov-directories-channel-search-block) > .views-exposed-form.localgov-search-channel-secondary {
   display: none;
--- a/src/Plugin/Block/ChannelSearchBlock.php
+++ b/src/Plugin/Block/ChannelSearchBlock.php
@@ -132,6 +132,7 @@ class ChannelSearchBlock extends BlockBase implements ContainerFactoryPluginInte
       ->disableRedirect();
 
     $views_exposed_filter_form = $this->formBuilder->buildForm(ViewsExposedForm::class, $form_state);
+    $views_exposed_filter_form['#id'] .= '-in-a-search-block';
     return $views_exposed_filter_form;
   }
 


finnlewis:
@ekes can you look at this?

ekes:
So I'm trying to jog my memory here. #52 #53 We had an attachment that was the map, which didn't take the arguments correctly, so it was changed to an embed.

The additional search field is coming from the embed?

I'm concerned that there are two and they are being visibly hidden. Usually they'll end up with the same CSS ID for example which can be problematic. So just checking if it's better to get rid of any that shouldn't be present - and then how...

Adnan-cds:
> We had an attachment that was the map, which didn't take the arguments correctly, so it was changed to an embed.
> The additional search field is coming from the embed?

That's correct.

> I'm concerned that there are two and they are being visibly hidden. Usually they'll end up with the same CSS ID for example which can be problematic.

Good point.  I missed that while adding the search **block** in September 2020.  In other words, this situation existed before the attachment was replaced with the map embed.

Here's the form id situation:
- 1x views-exposed-form-localgov-directory-channel-embed-map
- 2x views-exposed-form-localgov-directory-channel-node-embed

> So just checking if it's better to get rid of any that shouldn't be present - and then how...

Dropping the exposed forms may break the search.  Altering the form id within the **Search block** in the sidebar would be a [lot easier](https://github.com/localgovdrupal/localgov_directories/blob/1c67b233a3f770fa1a121e48a1e8470e255f4364/src/Plugin/Block/ChannelSearchBlock.php#L134).  I haven't tried this though.

ekes:
I'll have to do a fresh install of this to see what we get; but my gut feeling here is there should _either_ be one block enabled by default; or an exposed search field on the text view; the map doesn't need one as well.

This is all config, so it should also only be applied to sites installing the module for the first time; after that whatever search field options they go for should remain.

stephen-cox:
The problem here is we have an exposed search filter on the view and we have a block plugin for the exposed search filter so it can be displayed in the left-hand sidebar.

It's not immediately obvious how to remove the exposed search filter from the main page while keeping the block in the sidebar with just configuration.

One option to do this properly (rather than just hiding it with CSS) would be to change the ID of the form provided by the block plugin (as mentioned by Adnan) and then `hook_form_alter` the other exposed for so that it isn't displayed. Not sure how easy this will be though.

Adnan-cds:
I have resolved the duplicate Id issue.  I haven't tried to entirely remove exposed filter forms.  I feel that can be attempted as part of a different refactoring ticket.  I feel that way because the current approach of hiding the exposed filter forms (only when a search block is present) has been in use for a while.

stephen-cox (approved):
Given we're already hiding blocks with CSS, I'm happy to approve this as long as an issue is created to fix the fundamental problem of exposed view filters appearing in both the sidebar and the main content regions.

Adnan-cds:
Thanks Steve.  I have added a new ticket about refactoring #76 

Merging...

PR #83, opened 2021-05-07
Update composer dependencies to non-alpha localgov modules.

ekes:
Further to localgovdrupal/localgov#200 tidy up composer dependencies to the now existing 1.x versions.


andybroomfield (approved):

finnlewis:
@ekes you can merge this one!

PR #105, opened 2021-06-01
Remove rather than hide search #93.

### LINKED ISSUE #76: Refactor how identical exposed filter forms are hidden
The Directory channel pages and Directory pages end up with identical exposed filter forms when a search block is present in the sidebar. Currently we are hiding these redundant forms using CSS.  There has been [a proposal to entirely remove unneeded forms](https://github.com/localgovdrupal/localgov_directories/pull/74#issuecomment-829235007) using form_alter() or something similar.

### LINKED ISSUE #93: Search panel appears twice on Directory channel
See https://localgovdrupal.agile.coop/adult-health-and-social-care/money-employment-and-training-directory

The search panel should appear in the left hand column, not in the main column

ekes:
closes #76 closes #93


Adnan-cds:
Hi Ekes,
Is this ready for review or is it still in progress?  I am asking because it doesn't seem to have any effect :(

ekes:
Hmm.. still works here. What's the difference.

If you switch devel on and replace `removeExposedFilter` with 
```
>>  public static function removeExposedFilter(Markup $markup, array $render) {                                                                                                                                     
      // Sure there must be a better way in the pre_render to stop it adding the                                                                                                                                    
      // form, while accepting the parameters. But this does the same later.                                                                                                                                        
~     $count = 0;                                                                                                                                                                                                   
+     $output = $markup::create(preg_replace('|<form.*?class="[^"]*views-exposed-form.*?>.*?</form>|s', '', $markup, 1, $count));                                                                                   
>>    dsm($count);                                                                                                                                                                                                  
+     return $output;                                                                                                                                                                                               
    }   
```
what's the count?

If nothing is output then I guess caches or you aren't using the embedded map from the module but placing it some otherwise.
If '0' then I'm guessing the markup is different... which is the risk here. Maybe dump $markup->_toString() in that case?

ekes:
Thinking about it, if it is run, we can probably simplify the preg to just remove any form: `<form.*?>.*?</form>`. But should check that the cause isn't that it's not being run for some reason.

Adnan-cds:
##Sorry Ekes.  It does work :)  I got confused because we are still left with [two search forms](https://github.com/localgovdrupal/localgov_directories/blob/44c7048ac4d646c359c6267fa000315939a91e1e/src/DirectoryExtraFieldDisplay.php#L174) once you take out the CSS for hiding the duplicate search forms:
```
diff --git a/localgov_directories.libraries.yml b/localgov_directories.libraries.yml
index abf92b8..ad94650 100644
--- a/localgov_directories.libraries.yml
+++ b/localgov_directories.libraries.yml
@@ -10,7 +10,5 @@ localgov_directories_search:

 search_block:
   css:
-    state:
-      css/search-block.css: {}
     theme:
       css/search-block-style.css: {}
```
Screenshot:
screenshot-directory-channel-page

stephen-cox (approved):
It works. Let's merge it

ekes:
Ah!

> ##Sorry Ekes. It does work :) I got confused because we are still left with two search forms once you take out the CSS for hiding the duplicate search forms:

So it's there are three search forms prior to patching.

And we probably want to remove the search field from the list as well as the map. If and only if, the search block is enabled, or anyway. Probably anyway.

ekes:
> Not able to create a directory venue as map UX has disappeared

Editors not able to create/edit geo entities until permission granted.

ekes:
@Adnan-cds That make more sense? Removes from both embedded views (but does offer an alternative to show it one the list one should someone not want to use the block).

Adnan-cds:
> we probably want to remove the search field from the list as well as the map. If and only if, the search block is enabled
> That make more sense?

Makes total sense Ekes.  I couldn't have phrased it better :)

If it proves complicated to drop these search fields only when the search block is enabled, then I don't mind if you get rid of the search fields forever and provide some guidance in the README for those who don't want the search block.

ekes:
> If it proves complicated to drop these search fields only when the search block is enabled

Went with two 'extra fields' one disabled by default includes the search box, the other (the one enabled on sites already) has it disabled.

Adnan-cds:
Thanks Ekes, that should do.  I will do one final test tomorrow and let you know how it goes.

Adnan-cds:
Good morning Ekes,
I have tested it and all works fine in the Directory channel page.  That almost covers #93.  If you can please add a sentence in the README file saying a Channel search field is available for use as well, that may help the site builders experience.

On the very first comment above, you wanted to close #76 which involves making a similar change to Directory pages.  At the moment, I can see that the **Directory search** pseudo field is [enabled](https://github.com/localgovdrupal/localgov_directories/blob/453d6c89dca003a08565f301be50ab5a86034fbd/modules/localgov_directories_venue/config/install/core.entity_view_display.node.localgov_directories_venue.default.yml#L121) in **Directory venue** content type's *Default* display.  Is that intentional?  I mean, is this something site builders should enable or disable depending on the presence of the Search block?

finnlewis:
@ekes @Adnan-cds I'll have a quick look at adding the line in the README, once I understand what this is about. 
It would be good to get this merged!

finnlewis:
@Adnan-cds is this what you mean? https://github.com/localgovdrupal/localgov_directories/pull/105/commits/26072a5cb99a2fa314e77773c2f7735f72917585

@ekes have I written the truth?

finnlewis (approved):
I'm happy that I now understand this and have tested locally. I've also added to the README as requested by @Adnan-cds, so hopfully we can approve this and merge the fix.
I've also asked @andybroomfield in case he has more time to review than @Adnan-cds

Adnan-cds:
> I've also asked @andybroomfield in case he has more time to review than @Adnan-cds

I am going through it now.  Sorry for all the reviewing delays. @finnlewis @andybroomfield

Adnan-cds:
Additions to the README looks good Finn.  Thanks a lot.

The Directory **venue** pages still have two search forms.  I have attached a screenshot below.  Is this something we should fix in a different ticket?
screenshot-directory-venue-page

ekes:
@Adnan-cds 
> The Directory venue pages still have two search forms. I have attached a screenshot below. Is this something we should fix in a different ticket?

Is that not just the block and the field being displayed? 
Like you can just not display the block in the block config; or not display the field in the content type display fields config.

If so different issue I'd say, and just default config requiring only one, or the other, to be shown.

Adnan-cds:
Good morning :)

> If so different issue I'd say, and just default config requiring only one, or the other, to be shown.

Perhaps hide the *localgov_directory_search* extra field from the localgov_directories_venue content type's **default** display and mention in the README that this extra field is available for use in the localgov_directories_venue content type if need be.  

This is such a small change, I feel the time it will take you to raise another ticket is enough to change this as part of this pull request!  So perhaps change it here? :) @ekes @finnlewis

ekes:
> This is such a small change, I feel the time it will take you to raise another ticket is enough to change this as part of this pull request! So perhaps change it here? :) 

Actually seems like it should have been done when ever it was changed to default display them in the theme; but anyway. Added here (both Venue and Page); and added a note in the README.

@Adnan-cds Does that roll out of the box correctly for you now?
 * Single map on Directory Channel (and search box and facets as appropriate).
 * Single map on Venue (displayed as a content field)
 * Single search box on Venue and Entry

Adnan-cds (approved):
> Does that roll out of the box correctly for you now?

All good.  Thanks a lot :)

PR #109, opened 2021-06-02
Add title field for sort, remove some 'stopwords' from the sort.

### LINKED ISSUE #57: Sorting order of Directory entries
Hi Ekes,
I am trying to understand if the lack of a **Sorting criteria** on the Directory channel view (machine name: localgov_directory_channel) is a feature or a bug.  Could you please confirm.

I was about to add a Sorting criteria based on page Title on this View.  But then wondered if the Search API is supposed to dictate the ordering.  [Example URL](https://new.croydon.gov.uk/schools-and-education/schools/find-school-near-you/secondary-schools) of a Directory channel page.

FAO @ekes @finnlewis @willguv

ekes:
Adds a filter before indexing the (now added) title field that remove 'The' 'A' etc. (configurable). This field then used for the alphabetical sort of the items.

Unit Test failing locally; but rumour has it might work... Let's see.

### DIFF
--- a/config/install/search_api.index.localgov_directories_index_default.yml
+++ b/config/install/search_api.index.localgov_directories_index_default.yml
@@ -7,8 +7,9 @@ dependencies:
     - search_api.server.localgov_directories_default
     - core.entity_view_mode.node.directory_index
   module:
-    - search_api
     - node
+    - search_api
+    - localgov_directories
 id: localgov_directories_index_default
 name: Directories
 description: ''
@@ -33,6 +34,14 @@ field_settings:
         'entity:node':
           localgov_directories_page: directory_index
           node: directory_index
+  title:
+    label: Title
+    datasource_id: 'entity:node'
+    property_path: title
+    type: string
+    dependencies:
+      module:
+        - node
 datasource_settings:
   'entity:node':
     bundles:
@@ -50,10 +59,36 @@ processor_settings:
     all_fields: true
     fields:
       - rendered_item
+      - title
     weights:
       preprocess_index: -20
       preprocess_query: -20
   language_with_fallback: {  }
+  localgov_string_stopwords:
+    all_fields: false
+    fields:
+      - title
+    stopwords:
+      - a
+      - an
+      - and
+      - for
+      - if
+      - in
+      - into
+      - of
+      - 'on'
+      - or
+      - s
+      - t
+      - that
+      - the
+      - this
+      - to
+      - was
+    weights:
+      preprocess_index: -5
+      preprocess_query: -2
   rendered_item: {  }
   stemmer:
     all_fields: true
@@ -112,6 +147,7 @@ processor_settings:
     all_fields: true
     fields:
       - rendered_item
+    ignored: ._-
     spaces: ''
     overlap_cjk: 1
     minimum_word_size: '3'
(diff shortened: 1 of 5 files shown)


ekes:
OK it is finding the class there; and a few meaningful fails.

Oooh! If the name is just _THE_ should it actually not remove the only word? By rights of the general application it should, but by logic of the use it shouldn't.

ekes:
Actually thinking about testing it again, I come back to that a human should decide if it is: "Wateringhole, the" or "The Wateringhole" but...

ekes:
@willguv Are there really never any cases where you think that 'The' or similar is actually part of the name for the purpose of sorting?

willguv:
@ekes it's more about what letter a user would look under

[In Croydon](https://www.croydon.gov.uk/schools-and-education/schools/find-school-near-you/primary-infant-and-junior-schools?src=D7), my hunch is users would look under "W" for "The Woodside Academy"

My kids go to a school called "The Cavendish School" but everyone calls it "Cavendish"

This seems to be a common problem for schools

I looked on Google Maps too. Even names like "The Cotswolds" would be better placed under "C" not "T"

ekes:
I know this is for the sort order; but if you remove The you always remove _THE_. Hence my question. I assume you _do_ leave it if the name is only _THE_, or if it's an abbreviation you'd have to write it 'T.H.E.' or 'THE.' but what other rules do you need? The old classic is 'The Who' should be under T, so I asked myself - as this is automated - if there is never a case where the accepted name includes 'The ...' And Equally if there aren't lists that include other works 'A' 'For' 'To' that would not want to be used for the sort order. It just became obvious again when I was thinking of writing the tests with all the edge cases that it'll run into.

willguv:
@ekes I'm slightly confused here, sorry

 Are you proposing that we remove the word "The" completely from the list view? 

ie I'm scrolling down the list to "C" and see an entry called "The Cavendish School"

Or would "The" be removed completely, and I'll just see "Cavendish School"

ekes:
It's the 'field' (hidden automated not editable) used for sorting. Not for display. So it would still be 'The Cavendish School' in the list, but there would be 'Cavendish School' in the sort index so be under 'C'. That works for the use case so far I understand.

The questions are about what should be stored in this sort index field when the title is just 'THE' (maybe it's a abbreviation). What happens if a content editor that wants the 'THE' to be used for the sort (because it's part of the accepted name 'The Who' or it's the abbreviation at the start of the name) - they can't override it, it's an on-or-off thing.

The test above failed because 'The' on its own didn't get removed. Then I thought about it and realized if that's the string you probably _don't_ want to remove it. So it's list all the edge cases so it can be tested and still makes sense to do automatically.

Then if there are other words 'A' 'For' 'From' that shouldn't be used for the sort?

willguv:
@ekes do you have 15 mins to chat about this? 2.15?

Easier to do in person :)

ekes:
I've put a note about all the options in play https://github.com/localgovdrupal/localgov_directories/issues/57#issuecomment-853055891

So this can be draft till that's decided.

ekes:
Closing this in favour of starting afresh with the new approach and tasks decided on https://github.com/localgovdrupal/localgov_directories/issues/57#issuecomment-883244217

Issue #73, opened 2021-03-10
Configure Search API DB to deliver more useful results


willguv:
Croydon Hotjar data suggests that search takes 10% of the clicks on a typical landing page, but we've anecdotal evidence that the results aren't as good as they could be

Examples:

"aero drome" won't find "aerodrome"

"children's" won't find "children"

## Discussed with @stephen-cox and @ekes on 21 May

Search API DB allows basic fixes for:

- stemming/ plurals
- ignoring apostrophes
- ignoring spaces

Let's also check what's being indexed to give us the best results we can

##Later

Let's write something about the limitations of the built in search, and how it could be extended with a separate search service. Could talk to Waltham Forest as they're experimenting with Solr

willguv:
Search API DB allows basic fixes for

stemming/ plurals
ignoring apostrophes
ignoring spaces

Let's also check what's being indexed to give us the best results we can

ekes:
> "children's" won't find "children"

Test confirms that with the default configuration the body text, or title, including _children's_ is returned for the search _child_ and _children_.
And vice versa _children's_ and _childrens_ search find _children_.

> "aero drome" won't find "aerodrome"

_Aero_ and _Aerodrome_ and _Aero Drome_ found _Aerodrome_; _Drome_ alone doesn't.

Could the examples be an issue with a specific site configuration, are these string in different fields, or searched with "" marks? Are there other examples?

There are however additional strings included in the search (like the titles of fields) and it should be checked if all the fields are correctly displayed, so I'll make a PR for those.

willguv:
Thanks @ekes. This issue was raised in a Croydon content design meeting a while ago, and I think @benhillsjones mentioned it. 

Ben and I are catching up today and I'll raise it then. Or Ben, you can add extra info here if you have it

andybroomfield:
Closing as the merged PR should fix some of this. We have other open issues for improving search.

PR #110, opened 2021-06-02
Add and correct view modes for rendering pages and venue for search indexing

ekes:
Pages was using the default template, which could have been causing content to get missed.
Venue was good, but included the titles of the fields that's just repeated words not related to the content, and rendered some things unnecessarily.

Helps #73


ekes:
Oh! Is this still not a thing. Adding @stephen-cox @Adnan-cds @andybroomfield for review then!

Sorry I was just going through open PR's and saw this was sitting here.

modules/localgov_directories_venue/config/install/core.entity_view_display.node.localgov_directories_venue.directory_index.yml
```diff
@@ -1,3 +1,4 @@
+uuid: 4d6ed028-e23c-4552-b9fc-f9103df429fa
```
stephen-cox: Don't think this should be here

modules/localgov_directories_venue/config/install/core.entity_view_display.node.localgov_directories_venue.directory_index.yml
```diff
@@ -18,9 +19,10 @@ dependencies:
     - node.type.localgov_directories_venue
   module:
     - link
-    - telephone
     - text
     - user
+_core:
+  default_config_hash: yuSNmUeMHlWYAbP3Wezm7Ijgp8IkvQa9TREh5dYZ-qI
```
stephen-cox: Should be remove as well

stephen-cox (changes requested):
Couple of IDs to remove, but otherwise looks fine just from reviewing the config.

ekes:
So changes are now only hiding the labels from the exisiting indexd view mode for venue, and adding the equivalent to the page.

Adnan-cds:
I am happy for it to be merged once Steve approves it.

stephen-cox (approved):
I'm happy - let's merge

finnlewis (approved):
This looks good to me.

PR #113, opened 2021-06-02
Don't add dropdown if only one channel. #94

### LINKED ISSUE #94: Channel dropdown issue on Directory venue and Directory page
On https://localgovdrupal.agile.coop/adult-health-and-social-care/money-employment-and-training-directory/get-online-home and https://localgovdrupal.agile.coop/localgov-drupal-collaborators/annertech the search panel in the left hand column includes a channel drop down menu

eg

Screenshot 2021-05-20 at 18 10 43

This drop down menu should only appear in the page or venue is included in multiple channels. Both of the examples are only in one channel, so the drop down menu should not appear.

Separately I'm not convinced by this pattern, and we should include in user testing. For now, we need to fix the issue.

ekes:
Simple fix to #94 

Question. Now the single channel items don't have a dropdown but a title.

Screenshot_2021-06-02 A-library Drush Site-Install

The title is 'Search Libraries (optional)' with Libararies and (optional) in two different `<spans>` with different classes attached.  Title of the directory is now smaller and separte from Serach as the CSS in _form.sass (label > span) in LocalGov Theme. I don't think this module should include CSS to try and override that? That doesn't make sense. So it should either be accepted as is, or the CSS in LocalGov Theme should be more specific?


js/localgov_directories_entry_search.js
```diff
@@ -8,30 +8,33 @@
...
-        // and keeps the value of the selectors correct for the search
-        // they are on.
-        label.childNodes[0].addEventListener('change', function() {
-          var previousId = label.id.slice(0, -9);
-          var previous = document.getElementById(previousId);
-          previous.style.display = 'none';
-          var selected = document.getElementById(label.childNodes[0].value);
-          selected.style.display = 'block';
-          label.childNodes[0].value = previousId;
+      if (form_ids.length > 1) {
```
Adnan-cds, line 13: Err.. I would just do 
```
if (form_ids.length === 1) {
  return;
}
```
and then leave the rest of the code undisturbed to avoid adding more indentation to the main logic.  Not a big deal.

Adnan-cds (approved):
Works for me.  Thanks :)

js/localgov_directories_entry_search.js
ekes: The early returns debate :) 

Might be more readable if it's obvious here, but as this has been waiting to merge for ages I'm just going to do it.

Issue #100, opened 2021-05-24
"Back to search results" link [now repeats?]


willguv:
The original Croydon design for directories included a "back to search" results link above the search box

![directory-entry-top](https://user-images.githubusercontent.com/708140/119364251-9ed0a900-bca6-11eb-9d00-f48fe8ef9937.png)

This is shown if a user has reached this page from a directory channel. The link takes them back to the channel with the same facets applied as when they left it

If the user arrived by other means (eg via a Google search) the "back to search results" link is not shown

ekes:
Basically just the browsers back button then.

You can duplicate the browsers back button with javascript simply, it should be possible to restrict this to be displayed if the path looks like one that was the directory landing page.

willguv:
@ekes yep exactly that

ekes:
Seems this might sometimes repeat - looking at demo - 
![image](https://user-images.githubusercontent.com/237186/132840209-5f158b09-a5a6-435b-a6f3-00558d4d72e2.png)
I'd say https://github.com/localgovdrupal/localgov_directories/blob/42fc205fbf34213e534cb012e2547c39c9f9f1f8/js/localgov_directories_entry_search.js#L44 needs a `once`

PR #116, opened 2021-06-02
Add back to search results link #100

ekes:


js/localgov_directories_entry_search.js
```diff
@@ -32,6 +32,15 @@
           label.childNodes[0].value = previousId;
         });
       });
+
+      // Add a bact to search results page if the referrer is the same path minus the last /page.
+      if (document.URL.substr(0, document.URL.lastIndexOf('/')) == document.referrer.split('?')[0]) {
+        var searchForm = document.getElementById('block-localgov-directories-channel-search-block');
```
markconroy: `querySelector` is a more modern approach here, and also not using `var` as that can be reassigned and cause issues. Maybe change to:

`const searchForm = document.querySelector('#block-localgov-directories-channel-search-block')`

markconroy (commented):
This looks fine to me.

finnlewis (approved):
@ekes I think we can merge this one.

PR #120, opened 2021-06-11
Added search index and search result display modes

stephen-cox:
Part of https://github.com/localgovdrupal/localgov/issues/259


finnlewis (approved):
Not tested locally, but tests pass and it looks trivial enough to approve. 👍

PR #122, opened 2021-06-22
Feature/2.x/95 default permissions

ekes:
Adds default permissions for localgov drupal roles.


ekes:
Guess we want a new version of localgov_core that includes https://github.com/localgovdrupal/localgov_core/pull/87 (rather than changing dependencies here).

stephen-cox:
Changes coming to the workflow module add a contributor role so we will likely want to review the permissions assigned here and add the contributor role before approving.

Spec for changes: https://github.com/localgovdrupal/localgov_core/issues/92
PR for changes: https://github.com/localgovdrupal/localgov_core/pull/99

ekes:
So there's a 2.0.2 release of localgov_core that includes the updates to localgov_roles but `Error: Class 'Drupal\localgov_roles\RolesHelper' not found` so the tests are failing on that not being there still.

finnlewis:
@stephen-cox could you take a look?

stephen-cox:
To fix these tests we need a 2.1.2 release of the profile that include localgov_core 2.1.0 as they're now failing on the missing contributor role that is part of core 2.1.0

stephen-cox:
Tests now passing so this is ready for review

finnlewis (approved):
Looks good to me, tested locally too, works as expected. 
Tests pass too, which is always nice! 
Thanks @ekes and @stephen-cox !

PR #137, opened 2021-07-20
Release 2.0.2

stephen-cox:
Changelog:

#113 @ekes @Adnan-cds Don't add dropdown if only one channel.
#116 @ekes @markconroy @finnlewis Add back to search results link.
#122 @ekes @stephen-cox @finnlewis Add default permissions for LocalGov roles and workflows.


stephen-cox:
All these changes have been approved previously and the release is needed to fully deploy the workflow stuff, so merging and releasing.

PR #136, opened 2021-07-20
Feature/2.x/261 open referral

ekes:
- [ ] Requires localgovdrupal/localgov_openreferral#14
 - [ ] Also #138

PR #143, opened 2021-09-04
Feature/2.x/localgov geo browser

ekes:
Depends on (blocked by) localgovdrupal/localgov_geo#29
When the update to Geo and tagged release, this PR should have the localgov_geo version requirement bumped, and then merged.

Required for localgovdrupal/localgov_directories#136 (ie that PR is blocked by this).


composer.json
```diff
@@ -11,7 +11,7 @@
         "drupal/search_api": "^1.17",
         "drupal/search_api_autocomplete": "^1.3",
         "localgovdrupal/localgov_core": "^2.1",
-        "localgovdrupal/localgov_geo": "^1.0"
+        "localgovdrupal/localgov_geo": "dev-feature/1.x/area-and-ux-autocomplete"
```
Adnan-cds: Hi Ekes, because of this line, I am assuming you are still working on this and don't want us to review it yet.  But please clarify if I am wrong. @ekes

ekes: They all need to be released at the same time - so dependency version numbers will be bumped up one for each to the equivalent of the merged branch when those version exist. So they all need reviewing together. The localgov_geo PR is the main one for review (with guidance). There is an ovearching one that installs all the branches for testing together in https://github.com/localgovdrupal/localgov/pull/335

finnlewis (approved):
I'm happy with this in principle, but @ekes will need to edit the referenced branches in composer.json

PR #146, opened 2021-09-09
Fix dependencies for the 1.x branch to keep localgov_geo at 1.0.x rat…

### LINKED ISSUE #145: 1.x localgov_directories dependencies on localgov_geo
From localgovdrupal/localgov_alert_banner#172 (comment)

1.x events and directories shouldn't be downloading 1.1.x geo; they should have their version at 1.0.x
It's only the test that fails - so they could run 1.1.x - but... so if they were set so composer could download 1.x.x it should be more specific?

finnlewis:
…her than allow it to go to 1.1.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "drupal/search_api": "^1.17",
         "drupal/search_api_autocomplete": "^1.3",
         "localgovdrupal/localgov_core": "^1.0",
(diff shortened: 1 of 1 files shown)


andybroomfield (approved):
Locking geo to 1.0.x because 1.x core doesnt have the role

PR #152, opened 2021-09-23
Add field to sort by title for #57

ekes:


Adnan-cds:
RIP (Review in progress)

localgov_directories.install
```diff
@@ -84,3 +87,154 @@ function localgov_directories_update_8002() {
...
+  return implode(' ', $result);
+}
+
+/**
+ * Add new sort field to search index.
+ */
+function localgov_directories_update_8005() {
+  // Add to the search index.
+  $index = Index::load('localgov_directories_index_default');
+  if (!$index->getField('localgov_directories_index_default')) {
```
Adnan-cds: Hi Ekes, is this line (i.e. line 182) okay?  I have a feeling you meant to write ```$index->getField('localgov_directory_title_sort')```.

localgov_directories.module
```diff
@@ -467,3 +467,21 @@ function localgov_directories_remove_block_from_content_type_sidebar(string $blo
...
+ * Implements hook_search_api_index_items_alter().
+ *
+ * Populate localgov_directory_title_sort if empty.
+ */
+function localgov_directories_search_api_index_items_alter(IndexInterface $index, array &$items) {
+  foreach ($items as $item) {
+    if ($field = $item->getField('localgov_directory_title_sort')) {
+      $sort_value = $field->getValues();
+      if (empty($sort_value) || empty($sort_value[0])) {
+        // If the field is empty use the item title.
```
Adnan-cds, line 481: That's clever :)

localgov_directories.install
```diff
@@ -84,3 +87,154 @@ function localgov_directories_update_8002() {
...
+    $sort_title_field = new Field($index, 'localgov_directory_title_sort');
+    $sort_title_field->setLabel('Title (sort)');
+    $sort_title_field->setDatasourceId('entity:node');
+    $sort_title_field->setType('string');
+    $sort_title_field->setPropertyPath('localgov_directory_title_sort');
```
Adnan-cds, line 187: We may need to set the dependency here.  Something like the following.  If you please have a look.
```
$sort_title_field->setDependencies([
  'config' => [
    "field.storage.node.localgov_directory_title_sort",
  ],
]);
```

Adnan-cds (commented):
Code looks mostly okay.  I have pointed out two potential minor issues.  I will now test it on the Croydon site and get back to you.

localgov_directories.install
ekes: Nope. It just skips adding the field if there is already a field of the same name already there.

localgov_directories.install
ekes: Can you run this, and do a `drush cex` and check but I'm pretty certain it did this itself when I tested.

localgov_directories.install
Adnan-cds: The new field name is *localgov_directory_title_sort*.  But ```$index->getField('localgov_directories_index_default')``` is checking for the presence of *localgov_directories_index_default*.  That's why I thought I will point it out.  Perhaps I have missed something.

Adnan-cds: If you can please enlighten me on this Ekes.

localgov_directories.module
```diff
@@ -467,3 +467,21 @@ function localgov_directories_remove_block_from_content_type_sidebar(string $blo
...
+        $sort_value = [$item->getOriginalObject()->getEntity()->label()];
```
Adnan-cds: Any chance you can please pass the ```$sort_value``` through ```trim()```?  I have found a Directory venue page in our site whose title has a space at the beginning.  Possibly entered by mistake!  It is sitting at the top of the list as a result.

localgov_directories.install
Adnan-cds: You are right.  The dependency is automatically added.  Ignore what I have said.

Adnan-cds (commented):
Hi Ekes,
I have finished testing this and it works.  Thanks a lot :)

I have left a few comments in the code.  If you can please have a look.

Adnan-cds:
Forgot to mention that I had to **reindex** the Directory pages from /admin/config/search/search-api/index/localgov_directories_index_default after the update.  Can this be added as a return message for ```localgov_directories_update_8006()``` to save any confusion.

localgov_directories.install
ekes: Nah, just didn't see the difference for looking. Well saw the difference of `!` and not the longer string `localgov_directories_index_default`

ekes: Should be corrected now.

ekes:
> Forgot to mention that I had to **reindex** the Directory pages from /admin/config/search/search-api/index/localgov_directories_index_default after the update. Can this be added as a return message for `localgov_directories_update_8006()` to save any confusion.

Popped it in 8005 as this is where the field is added and a reindex requested. Assume the wording is clear enough?

finnlewis:
@ekes is just checking the necessity of ```trim()```  before merging.

localgov_directories.module
ekes: So I was thinking about this.

My immediate reaction was we should then trim the field itself, rather than the default value entered into the field,
*but* then I thought if you are entering a string for sorting that's really the value it should be sorted on with exactly what was entered,
so then trim()ing the automated output starts to make sense, so I'll add that.

ekes:
Latest push adds the trim() see thoughts on the commit.

Also merges with updates from 2.x branch.

Should all still look good to merge now?

localgov_directories.install
Adnan-cds: Thanks Ekes :)

Adnan-cds:
> Popped it in 8005 as this is where the field is added and a reindex requested. Assume the wording is clear enough?

+1

> Latest push adds the trim() see thoughts on the commit.

Works great :+1: 

> Should all still look good to merge now?

Looking good to me.  Thanks a lot once again :)

finnlewis (approved):
Looks like @Adnan-cds has tested and reviewed code and is happy, which makes me even happier!

PR #159, opened 2021-10-04
Add optional support for localgov_scarfolk theme block placement out …

finnlewis:
…of the box.

### DIFF
--- a/config/optional/block.block.localgov_directories_facets.yml
+++ b/config/optional/block.block.localgov_directories_facets.yml
@@ -7,9 +7,9 @@ dependencies:
     - facets
     - node
   theme:
-    - localgov_scarfolk
-id: localgov_directories_facets_scarfolk
-theme: localgov_scarfolk
+    - localgov_base
+id: localgov_directories_facets
+theme: localgov_base
 region: sidebar_first
 weight: 5
 provider: null
--- a/config/optional/block.block.localgov_directories_channel_search_block.yml
+++ b/config/optional/block.block.localgov_directories_channel_search_block.yml
@@ -6,7 +6,7 @@ dependencies:
     - node
   theme:
     - localgov_base
-id: localgov_directories_channel_search_block_base
+id: localgov_directories_channel_search_block
 theme: localgov_base
 region: sidebar_first
 weight: 0
--- a/config/optional/block.block.localgov_directories_channel_search_block_base.yml
+++ b/config/optional/block.block.localgov_directories_channel_search_block_base.yml
@@ -6,7 +6,7 @@ dependencies:
     - node
   theme:
     - localgov_base
-id: localgov_directories_channel_search_block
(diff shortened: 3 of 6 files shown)


config/optional/block.block.localgov_directories_facets.yml
```diff
@@ -7,9 +7,9 @@ dependencies:
...
   theme:
-    - localgov_base
-id: localgov_directories_facets
-theme: localgov_base
+    - localgov_scarfolk
+id: localgov_directories_facets_scarfolk
+theme: localgov_scarfolk
 region: sidebar_first
 weight: 5
 provider: null
```
stephen-cox: This file should be renamed to  `config/optional/block.block.localgov_directories_facets_scarfolk.yml` to match the block ID.

stephen-cox (changes requested):
While installing this with the profile branch that enables the Scarfolk theme and the Demo content module I see the following error message repeated twice. 

```
[error]  Block localgov_directories_channel_search_block is missing.  Cannot update its visibility settings.
```

Looks like the `CHANNEL_SEARCH_BLOCK` constant needs to be updated as well.

https://github.com/localgovdrupal/localgov_directories/blob/4ceaf0b1f73dce5e536c400cc6b7096c1be31ae4/src/Constants.php#L14

config/optional/block.block.localgov_directories_facets.yml
finnlewis: Thanks @stephen-cox - I'd missed a commit and push. Please see now, added two new files for the localgov_scarfolk theme config.

finnlewis:
Thanks @stephen-cox I'd missed a commit, should be good now.

finnlewis:
Actually - what about the `CHANNEL_SEARCH_BLOCK` ? It seems to work without changing that now, as I've not changed the id of the original block.

finnlewis:
Looking at the code, I think the block id is unchanged, as 
`localgov_directories_channel_search_block` 
even if the config id changes.

Both configs reference the same actual block plugin: 

https://github.com/localgovdrupal/localgov_directories/blob/feature/158-scarfolk/config/optional/block.block.localgov_directories_channel_search_block.yml#L14

https://github.com/localgovdrupal/localgov_directories/blob/feature/158-scarfolk/config/optional/block.block.localgov_directories_channel_search_block_scarfolk.yml#L14

finnlewis:
Keen to merge this unless there are any objections.

finnlewis:
@ekes any thoughts on the above?

ekes:
> @ekes any thoughts on the above?

I can have a review of all that conditional config logic, and block placement, but @Adnan-cds probably already remembers it, so might be better placed for a quick answer.

stephen-cox (approved):
Looking good now

finnlewis:
Thanks @stephen-cox ! This is why two pairs of eyes are good!

PR #191, opened 2022-06-06
Reexport embedded map view with leaflet 2.2

ekes:
For https://github.com/localgovdrupal/localgov_geo/pull/59 (and so requires it).


andybroomfield (approved):
Ok, assuming this is just the new leaflet config.

Issue #198, opened 2022-06-24
Test error - The text "Page 1" was not found anywhere in the text of the current page.


stephen-cox:
```
Testing Drupal\Tests\localgov_directories_page\Functional\PathIntegrationTest
Path Integration (Drupal\Tests\localgov_directories_page\Functional\PathIntegration)
 ✘ Directory integration
   ┐
   ├ Behat\Mink\Exception\ResponseTextException: The text "Page 1" was not found anywhere in the text of the current page.
   │
   ╵ /app/vendor/behat/mink/src/WebAssert.php:811
   ╵ /app/vendor/behat/mink/src/WebAssert.php:262
   ╵ /app/web/core/tests/Drupal/Tests/WebAssert.php:898
   ╵ /app/web/modules/contrib/localgov_directories/modules/localgov_directories_page/tests/src/Functional/PathIntegrationTest.php:94
   ╵ /app/vendor/phpunit/phpunit/src/Framework/TestResult.php:726
   ┴
```

PR #200, opened 2022-06-30
Removed 'Block %block-id is missing. Cannot update its visibility settings' error message

### LINKED ISSUE #194: Remove `Block %block-id is missing.  Cannot update its visibility settings.` error message as it serves no purpose
When enabling directories in Microsites two separate error alerts are thrown:

```
 [error]  Drupal\search_api\SearchApiException while adding Views handlers for field Content » Directory channels on index Directories: Could not retrieve data definition for field 'Directory channels' on index 'Directories'. in Drupal\search_api\Item\Field->getDataDefinition() (line 482 of /app/web/modules/contrib/search_api/src/Item/Field.php). 
 [error]  Drupal\search_api\SearchApiException while computing Views data for index Directories: Could not retrieve data definition for field 'Directory channels' on index 'Directories'. in Drupal\search_api\Item\Field->getDataDefinition() (line 482 of /app/web/modules/contrib/search_api/src/Item/Field.php). 
```

Looks related to #43 and #157

```
[error]  Block localgov_directories_channel_search_block is missing.  Cannot update its visibility settings. 
```

Looks to be due to https://github.com/localgovdrupal/localgov_directories/blob/0d2d1e51458479c06339a47719d69c283c619649/localgov_directories.module#L229

stephen-cox:
This error message is being thrown when directories are being enabled in the Microsites distribution and doesn't convey any useful information, so let's remove it.

The function that produces the error message attempts to update the visibility settings of the `localgov_directories_channel_search_block` (which is the only time this function is used). The aim is to display the search block on Directory Pages, but we don't want this displaying on these pages, only Channels, so we don't want to fix the error.

Fixes #194

### DIFF
--- a/localgov_directories.module
+++ b/localgov_directories.module
@@ -392,10 +392,6 @@ function localgov_directories_add_block_to_content_type_sidebar(string $block_id
 
   $block_config = Block::load($block_id);
   if (empty($block_config)) {
-    Drupal::service('logger.factory')->get('localgov_directories')->error('Block %block-id is missing.  Cannot update its visibility settings.', [
-      '%block-id' => $block_id,
-    ]);
-
     return FALSE;
   }
 


stephen-cox:
Tests are failing due to #198

ekes:
Pretty certain the intention was that the search block is on the pages. It's Croydon's pattern https://www.croydon.gov.uk/schools-and-education/schools/find-school-near-you/secondary-schools/archbishop-tenisons-church-england-school comes in from there too https://github.com/localgovdrupal/localgov_directories/commit/a7d31ac0507f4753d9c74ee40475ae7cdd8ebd01

stephen-cox:
Looks like that was the idea, but doesn't look to be working any longer. Just done a fresh install on the test site and the search block isn't placed on directory items: https://test.localgovdrupal.org/localgov-drupal-collaborators/agile-collective

The question then becomes, is placing the search block for directory items still desirable for LocalGov and/or Microsites?

@msayoung Do we have wireframes for directory items in Microsites? Do we want a search block on pages and venues?

msayoung:
I am not a fan of having them there myself, as it doesn't make much sense to me. But probably needs testing with some end users  tbh.

On the wireframes for microsites we do not have them on the actual page / venue.

Adnan-cds:
> but doesn't look to be working any longer. Just done a fresh install on the test site and the search block isn't placed on directory items

It won't work for the Scarflok theme due to the differences in block ids:
- Expected: [localgov_directories_channel_search_block](https://github.com/localgovdrupal/localgov_directories/blob/2.x/src/Constants.php#L14)
- In case of Scarfolk: [localgov_directories_channel_search_block_scarfolk](https://github.com/localgovdrupal/localgov_directories/blob/2.x/config/optional/block.block.localgov_directories_channel_search_block_scarfolk.yml)

It works for other themes such as localgov_base.

stephen-cox:
I'm going to close this PR and move the discussion to whether block placement code is desirable to #203

PR #202, opened 2022-06-30
Add additional content types to search index datastore directly to the configuration entity

### LINKED ISSUE #201: It's not possible to enable Venues as part of a site install
From https://github.com/localgovdrupal/localgov_microsites_group/issues/108

Enabling directories during a site install using Drush hangs indefinitely. This is because the config change when adding new content types to the Search API data source for the second time triggers a content reindex, which hangs in Drush.

The code in `hook_ENTITY_TYPE_insert()` that adds the directory item content type to the search index data store is run twice during a site install; first for `localgov_directories_page` and then `localgov_directories_venue`. The second time the index configuration is saved a content reindex is triggered. When run through Drush this triggers a second Drush instance to run the batch process, which hangs. It works fine when running a site install through the browser and doesn't cause a problem enabling venues after a site install.

The code the causing the problem is: https://github.com/localgovdrupal/localgov_directories/blob/0d2d1e51458479c06339a47719d69c283c619649/localgov_directories.module#L176-L181

To fix this I think we'll need to edit the configuration directly, rather than through the Index entity.

FYI @finnlewis @ekes

stephen-cox:
Fixes #201

### DIFF
--- a/localgov_directories.module
+++ b/localgov_directories.module
@@ -173,13 +173,6 @@ function localgov_directories_field_config_insert(FieldConfigInterface $field) {
       return;
     }
 
-    $configuration = $datasource->getConfiguration();
-    $configuration['bundles']['default'] = FALSE;
-    if (!in_array($entity_bundle, $configuration['bundles']['selected'])) {
-      $configuration['bundles']['selected'][] = $entity_bundle;
-    }
-    $datasource->setConfiguration($configuration);
-
     $index_field = $index->getField('rendered_item');
     if ($index_field) {
       $configuration = $index_field->getConfiguration();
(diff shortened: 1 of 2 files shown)


stephen-cox:
The failing test is not to do with this PR #198

ekes:
https://www.drupal.org/project/search_api/issues/3294061

finnlewis:
@ekes will re-read this and consider implications.

stephen-cox:
This needs to be resolved before we can enable venues as part of a site installed with https://github.com/localgovdrupal/localgov_microsites_group/pull/135

finnlewis:
@ekes you wanted to review this before it gets merged I think.

ekes:
Discussion is on the issue #201 

Conclusion is switch the index off (and on again); and #206 also fixes the specific case of install profile installation.

stephen-cox:
Updated this to disable the index and enable it after config changes, but there's a test failure. This was the same one as before. I needed to resave the datastore after manually adding the configuration to it. I think I might undo the last commit as at least that worked and all tests passed

ekes:
So I've tested adding `localgov_directories` `localgov_directories_venue` to the localgov install profile. It also hangs.
However when you use #206 and add `localgov_directories` `localgov_directories_venue` `localgov_directories_db` it does not hang.

Problem solved?

Not for localgov_microsites profile. Or rather localgov_microsites_group_directories. Using #206 adding `localgov_directories_db` to it doesn't help. I assume because it installs the localgov_directories dependencies in localgov_microsites_group_directories. So the order is different.

However, you can consistently enable `localgov_directories_db` at the end by not adding it to the .info.yml but in
```
function localgov_microsites_install() {                                                                                                                                                                            
  \Drupal::service('module_installer')->install(['localgov_directories_db']);                                                                                                                                       
}                                                                                                                                                                                                                   
```
#206 then reliably works with localgov_microsites_group_directories enabling the directories modules including venue.

stephen-cox:
Thanks @ekes. Closing this as it's not longer required.

PR #205, opened 2022-07-07
Use the shared paragraph field in the localgov_paragraphs_layouts module

msayoung:


msayoung:
Requires: https://github.com/localgovdrupal/localgov_paragraphs/pull/62

PR #206, opened 2022-07-11
Initial move of server to seperate module.

ekes:
Working on the same principle as https://github.com/localgovdrupal/localgov_search/pull/34
Has the added bonus of assisting like https://github.com/localgovdrupal/localgov_directories/issues/201

Would need. Tests altering.
I'm thinking that existing tests would require localgov_directories_db.
There would be an additional Functional Base Test than any server
modules can inherit to check integration (as
localgovdrupal/localgov_search#34)


finnlewis:
This goes with https://github.com/localgovdrupal/localgov_search/pull/40

finnlewis:
From discussion in the search working group: 

@ekes thinks it would be good to write an update path to enable the module but not install it's config! 

@ekes will need to look at this before further testing.

modules/localgov_directories_db/localgov_directories_db.install
```diff
@@ -0,0 +1,45 @@
...
+function localgov_directories_db_install($is_syncing) {
+  if (
+    !$is_syncing &&
+    ($index = Index::load('localgov_directories_index_default')) &&
+    ($index->getServerId() == '')
+  ) {
+    $server = Server::load('localgov_directories_default');
+    $module_path = \Drupal::service('extension.list.module')->getPath('localgov_directories_db');
+    $index_solr_config = Yaml::decode(file_get_contents($module_path . '/config/conditional/search_api.index.localgov_directories_index_default.yml'));
+    $new_config_index = Index::create($index_solr_config);
```
Adnan-cds: Please reconsider variable naming of `$index_solr_config`.

modules/localgov_directories_db/config/conditional/search_api.index.localgov_sitewide_search.yml
```diff
@@ -0,0 +1,137 @@
+langcode: en
```
Adnan-cds: Do we really need the **search_api.index.localgov_sitewide_search.yml** file in the localgov_directories_db submodule?

Adnan-cds (commented):
Looks mostly okay.  I have added two minor suggestions.

It may be better to include in the README that the localgov_directories_db module needs to be enabled on a new site for any kind of Directory search to work.

Also, if any other search backend is later enabled and a module *similar* to localgov_directories_db is installed to support that backend, there could be a conflict over [the server setting](https://github.com/localgovdrupal/localgov_directories/blob/0dfb57652e8153053a8f2067b9da71c3ae7ccf2d/config/install/search_api.index.localgov_directories_index_default.yml#L65).  So may be better to uninstall localgov_directories_db before switching to a new backend.  The README can clarify all these.

modules/localgov_directories_db/localgov_directories_db.install
ekes: https://github.com/localgovdrupal/localgov_directories/pull/206/commits/815dee95ea65d45553fa3450b0a0494399d4c24b

modules/localgov_directories_db/config/conditional/search_api.index.localgov_sitewide_search.yml
ekes: https://github.com/localgovdrupal/localgov_directories/pull/206/commits/815dee95ea65d45553fa3450b0a0494399d4c24b

ekes:
> It may be better to include in the README that the localgov_directories_db module needs to be enabled on a new site for any kind of Directory search to work.

Something like
https://github.com/localgovdrupal/localgov_directories/pull/206/commits/31dcbee119f5926c96169bf3e33701f754ee6ca7
?

modules/localgov_directories_db/config/conditional/search_api.index.localgov_sitewide_search.yml
Adnan-cds: Thanks.

Adnan-cds (approved):
Looks good now.  Thank you :)

Issue #213, opened 2022-07-25
Missing Schema value for localgov_directories_db


andybroomfield:
On upgrading to localgov_directories 2.3.0 and running drush updb -y.

>  [warning] Schema information for module localgov_directories_db was missing from the database. You should manually review the module updates and your database to check if any updates have been skipped up to, and including, localgov_directories_db_update_8000().

andybroomfield:
This is resolved.

PR #214, opened 2022-07-26
Fix for #213

ekes:


ekes:
If someone can double check this. I just realized I'm not 100% sure that module exists is returning the correct value when it has just been enabled in the previous update, my db might not have been clean by that point.

andybroomfield (approved):
This resolves the issue for me.

ekes:
Same update as [search](https://github.com/localgovdrupal/localgov_search/pull/43#issuecomment-1195620755) to make sure it's good if your ran 8007 and 8008 in one go, as well as separately.

andybroomfield:
See https://github.com/localgovdrupal/localgov_search/pull/43#issuecomment-1196369312

PR #216, opened 2022-07-28
2.3.1 release

finnlewis:
* Fix for #213

* Fix if both 8007 and 8008 are run at the same time.

Running 8007 and then 8008 seperately works with the module check, but
not when both are run consecutively in the same call. Reseting module
list didn't help.


stephen-cox (approved):
Let's get it done!

PR #217, opened 2022-08-04
Fix pathauto patterns

ekes:
localgovdrupal/localgov_core#142


ekes:
Ah! These, their automatic --prefer-source checkouts will include the 2.x-release branch merge commits. Think that's fine?

finnlewis (approved):
Tested locally, thanks @ekes

PR #220, opened 2022-08-19
Add search result highlight to search result highlight display

### LINKED ISSUE #215: Search results: set consistent view mode [localgov_directories]
As part of reviewing the search functionality, it has become apparent that there is no consistent presentation of search results across the distribution.

We need to

- [ ]  ensure each content type has a search_result display node
- [ ]  configure the search_result display mode to have the appropriate fields

In the case of localgov_directories, we need to ensure appropriate configuration for

- [ ] localgov_directory
- [ ] localgov_directories_page
- [ ] localgov_directories_venue

andybroomfield:
Fix #215

This is for sitewide search, not the directory search.

Usual cavet about using search_api as dependency applies.
(If not using search_api, the search result display will be blank)

### DIFF
--- a/config/install/core.entity_view_display.node.localgov_directory.search_result.yml
+++ b/config/install/core.entity_view_display.node.localgov_directory.search_result.yml
@@ -15,20 +15,19 @@ targetEntityType: node
 bundle: localgov_directory
 mode: search_result
 content:
-  body:
-    label: hidden
-    type: text_summary_or_trimmed
-    weight: 0
-    settings:
-      trim_length: 600
+  search_api_excerpt:
+    settings: {  }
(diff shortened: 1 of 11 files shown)


ekes (approved):
Manually tested with https://github.com/localgovdrupal/localgov_search/pull/45 I'd say we want that in at the same time this is released.

finnlewis:
@ekes would like us to add a test for this, following the example on https://github.com/localgovdrupal/localgov_step_by_step/pull/64/files#diff-afe33a5fb745c57e43f66de12ec75bfae6c2526c9e584f3e522a73eeb73800e6

finnlewis (approved):

PR #223, opened 2022-09-08
Update the form display for the directory promo page

msayoung:


msayoung:
Before

image

After: 

image

- exposed the summary field (as per other Directory items)
- changed the banner field display, to not default to select seconday banner paragraph as it caused confusion in testing
- put the banner and standfirst fields in a fieldset with help text
- renamed "Content" field to "Layouts and rich content" - updated help text

msayoung:
@finnlewis @willguv 
Thoughts on the above changes? If you think they are good we can roll them out to Page, Events and News where we have the same. 

Particular changes:

- changed the banner field display, to not default to select seconday banner paragraph as it caused confusion in testing
- put the banner and standfirst fields in a fieldset with help text
- renamed "Content" field to "Layouts and rich content" - updated help text

Does it help? Is it clearer?

finnlewis:
Thanks @msayoung - all those changes look like good ones to me!

finnlewis (approved):
Tested locally, and works as advertised. The form is certainly tidier and I expect easier to use. 

Tests all pass and this is only for new installs. 

Thanks @msayoung !

PR #224, opened 2022-09-09
2.3.3 release

finnlewis:


markconroy (approved):
This looks really good to me.

stephen-cox:
While testing this I've found a couple of quite serious issues with Directories that look like they result from the creation of the localgov_directories_db module to handle config for Search API DB; #227 #228. As it stands, I don't think the 2.3.x branch is usable for fresh installs.

They don't look related to the changes here, just flagging that we need to address these issues with in this release or another release as soon as they're fixed.

finnlewis:
We're discussing this at our Merge Monday meeting.

We noted: 

1. There are some broken things that this release does not address. See #227 and #228
2. There is a coding standards 'fail' which we're happy to ignore for now: `ERROR: Ruleset /home/runner/work/localgov_directories/localgov_directories/html/phpcs.xml.dist is not valid`

We're happy to release the updates with the above notes.

PR #230, opened 2022-09-14
Update config to allow the primary banner as an option on the banner …

msayoung:
…field


modules/localgov_directories_promo_page/config/install/field.field.node.localgov_directory_promo_page.localgov_directory_banner.yml
```diff
@@ -21,6 +22,7 @@ settings:
   handler: 'default:paragraph'
   handler_settings:
     target_bundles:
+      localgov_banner_primary: localgov_banner_primary
       localgov_banner_secondary: localgov_banner_secondary
     negate: 0
     target_bundles_drag_drop:
```
stephen-cox, line 28: I don't think you need all the disabled `target_bundles_drag_drop` items; they are referencing paragraph types that may not exist.

I suggest removing all the items with `enabled: false`, shortening this to
```yaml
    target_bundles_drag_drop:
      localgov_banner_primary:
        weight: 0
        enabled: true
      localgov_banner_secondary:
        weight: 1
        enabled: true
```

stephen-cox (approved):
I've a suggestion to simplify the config, otherwise works fine

stephen-cox:
Just found something slightly strange with this. A Secondary Banner paragraph is added to the form when creating a new page and this looks a little odd (particularly at medium screen widths).
image

msayoung:
There is another PR removing the secondary banner as the default option 

https://github.com/localgovdrupal/localgov_directories/pull/223

also contains some other improvements.

msayoung:
I've made those improvements @stephen-cox

markconroy (approved):

PR #232, opened 2022-09-22
Add template to render directory facets as dropdowns

### LINKED ISSUE #231: Directory facets aren't rendered when configured as a dropdown
In the facet options for the Directory Facet it's possible to set the widget to a dropdown, but this renders a dropdown with only a 'Choose` option and a the word Facet Facet underneath.

image

There's no Twig template in the module for a dropdown facet and the one in the base theme doesn't handle dropdowns correctly. The directory facet needs handling slightly differently than a standard facet dropdown as a separate dropdown for each facet type should be rendered.

stephen-cox:
Allows directory facets to be configured as dropdowns.

Fixes #231

### DIFF
--- a/templates/facets-item-list--dropdown--localgov-directories-facets.html.twig
+++ b/templates/facets-item-list--dropdown--localgov-directories-facets.html.twig
@@ -24,26 +24,22 @@
  * @ingroup themeable
  */
 #}
-<div class="facets-widget- {{- facet.widget.type -}}">
-  {% if facet.widget.type %}
-    {%- set attributes = attributes.addClass('item-list__' ~ facet.widget.type) %}
-  {% endif %}
-  {% if items or empty %}
-    {%- if title is not empty -%}
-      <h3>{{ title }}</h3>
-    {%- endif -%}
+{%- set attributes = attributes.addClass('item-list__' ~ facet.widget.type) %}
+<div class="facets-widgets- {{- facet.widget.type -}}">
+  {%- if title is not empty -%}
+    <h3>{{ title }}</h3>
+  {%- endif -%}
 
-    {%- if items -%}
-      {%- for group in items -%}
-        <label>{{ group.title }}</label>
-        <ul{{ attributes }}>
-        {%- for item in group.items -%}
-          <li{{ item.attributes }}>{{ item.value }}</li>
-        {%- endfor -%}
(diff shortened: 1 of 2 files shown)


Adnan-cds:
Hi Steve,
When I switch to Dropdown for the Facet widget display, I don't see any Facet labels :(
screenshot-facet-dropdown-without-label

The label is actually there but [hidden by the facets](https://git.drupalcode.org/project/facets/-/blob/2.0.x/css/general.css) module.  Don't know why.

stephen-cox:
> When I switch to Dropdown for the Facet widget display, I don't see any Facet labels :(

I noticed this too. Not sure why the Facets module is intentionally hiding labels.

I have updated the Twig template so labels now display

templates/facets-item-list--dropdown--localgov-directories-facets.html.twig
```diff
@@ -0,0 +1,45 @@
+{#
+/**
+ * @file
+ * Default theme implementation for a facets item dropdown.
```
Adnan-cds: Any chance the accuracy of the above comment line can be improved?

Adnan-cds (approved):
Works for me now.  Thanks.

templates/facets-item-list--dropdown--localgov-directories-facets.html.twig
finnlewis: I think this is actually pretty consistent with the default facets templates. 
https://git.drupalcode.org/project/facets/-/blob/2.0.x/templates/facets-item-list.html.twig#L4
I couldn't think of anything better, what did you have in mind @Adnan-cds ?

finnlewis (approved):
Tested locally and works for me. I see the lables too. 
Thanks @stephen-cox and @Adnan-cds !

markconroy (approved):

templates/facets-item-list--dropdown--localgov-directories-facets.html.twig
Adnan-cds: It's the LocalGov flavour of the facets dropdown rather than the default one, isn't it?

stephen-cox: I've updated the comment to make it clear it refers to directory facets

PR #235, opened 2022-10-03
added 2 new facets plugins for processor and query_type. Together the…

### LINKED ISSUE #193: Use an 'and' query between facet groups and 'or' within them.
As the facets are dynamically generated, currently there is a choice in the facets setting for a single block as either 'or' (show results with either facet 1 or facet 2) or 'and' (show results with both facet 1 and facet 2).

However this can be confusing when there are mulktiple groups of facets.
We need to have the option of 'or' within the block, and 'and' between the blocks, whilst maintaining content editors abilitry to create facets.

Eg.  [SEND with over 18 and 16 - 18 checked](https://www.brighton-hove.gov.uk/directories/special-educational-needs-and-disability-send-support?f%5B0%5D=localgov_directories_facets%3A756&f%5B1%5D=localgov_directories_facets%3A3086) compared to [SEND with over 18 and help for child ticked](https://www.brighton-hove.gov.uk/directories/special-educational-needs-and-disability-send-support?f%5B0%5D=localgov_directories_facets%3A1466&f%5B1%5D=localgov_directories_facets%3A3086) the 'just right' entry is visible. Really the logic here should be (age:16 - 18 OR age:over 18) AND (help for:child).

Screenshot 2022-06-13 at 7 16 51 pm

It looks like we should be able to do this with a facet pre processer plugin.

snpower:
…y provide a new pre_query processor on facets to AND facet groups together while still allowing OR within a facet group. This is based on initial code provided by Andy Broomfield

### DIFF
--- a/src/Plugin/facets/processor/LocalGovDirectoriesProcessor.php
+++ b/src/Plugin/facets/processor/LocalGovDirectoriesProcessor.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Drupal\localgov_directories\Plugin\facets\processor;
+
+use Drupal\facets\FacetInterface;
+//use Drupal\facets\Processor\BuildProcessorInterface;
+use Drupal\facets\Processor\PreQueryProcessorInterface;
+use Drupal\facets\Processor\ProcessorPluginBase;
+
+/**
+ * ANDs LocalGov Directories Facet Groups while keeping OR within each group.
+ *
+ * @FacetsProcessor(
+ *   id = "localgov_directories_processor",
+ *   label = @Translation("LocalGov Directories - AND Facet Groups"),
+ *   description = @Translation("ANDs LocalGov Directories Facet Groups while keeping OR within each group."),
+ *   stages = {
+ *     "pre_query" = 35
+ *   }
+ * )
+ */
+//class LocalGovDirectoriesProcessor extends ProcessorPluginBase implements BuildProcessorInterface {
+class LocalGovDirectoriesProcessor extends ProcessorPluginBase implements PreQueryProcessorInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  //public function build(FacetInterface $facet, array $results) {
+   // return $results;
+  //}
+  public function preQuery(FacetInterface $facet) {
+    $active_items = $facet->getActiveItems();
+    $facet->setActiveItems($active_items);
+  }
+
+  /**
+   * {@inheritdoc}
+   *
+   * String corresponds to a key on the $query_types array as defined
+   * within hook_facets_search_api_query_type_mapping_alter().
+   *
+   * @see hook_facets_search_api_query_type_mapping_alter()
+   */
+  public function getQueryType() {
+    return 'localgov_directories';
+  }
+
+}
--- a/src/Plugin/facets/processor/LocalGovDirectoriesProcessor.php
+++ b/src/Plugin/facets/processor/LocalGovDirectoriesProcessor.php
@@ -39,4 +39,5 @@ class LocalGovDirectoriesProcessor extends ProcessorPluginBase implements PreQue
   public function getQueryType() {
     return 'localgov_directories';
   }
+
 }
--- a/src/Plugin/facets/processor/LocalGovDirectoriesProcessor.php
+++ b/src/Plugin/facets/processor/LocalGovDirectoriesProcessor.php
@@ -3,7 +3,6 @@
 namespace Drupal\localgov_directories\Plugin\facets\processor;
 
 use Drupal\facets\FacetInterface;
-//use Drupal\facets\Processor\BuildProcessorInterface;
 use Drupal\facets\Processor\PreQueryProcessorInterface;
 use Drupal\facets\Processor\ProcessorPluginBase;
 
@@ -19,15 +18,11 @@ use Drupal\facets\Processor\ProcessorPluginBase;
  *   }
  * )
  */
-//class LocalGovDirectoriesProcessor extends ProcessorPluginBase implements BuildProcessorInterface {
 class LocalGovDirectoriesProcessor extends ProcessorPluginBase implements PreQueryProcessorInterface {
 
   /**
    * {@inheritdoc}
    */
-  //public function build(FacetInterface $facet, array $results) {
-   // return $results;
-  //}
   public function preQuery(FacetInterface $facet) {
     $active_items = $facet->getActiveItems();
     $facet->setActiveItems($active_items);
@@ -44,5 +39,4 @@ class LocalGovDirectoriesProcessor extends ProcessorPluginBase implements PreQue
   public function getQueryType() {
     return 'localgov_directories';
   }
-
 }
--- a/src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
+++ b/src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
@@ -7,8 +7,7 @@ use Drupal\facets\Result\Result;
 use Drupal\search_api\Query\QueryInterface;
 
 /**
- * Provides support to "AND" facet groups while keeping the operator within a
- * facets as an "OR".
+ * "AND" facet groups while keeping the operator within a facets as an "OR".
  *
  * @FacetsQueryType(
  *   id = "localgov_directories_query_type",
@@ -47,6 +46,7 @@ class LocalGovDirectoriesQueryType extends QueryTypePluginBase {
           $bundle[$directory_facet->bundle()][] = $directory_facet->id();
         }
 
+        $filter = NULL;
         foreach ($bundle as $bundle_name => $group_items) {
           unset($filter);
           $filter = $query->createConditionGroup($operator, ['facet:' . $field_identifier . '.' . $bundle_name]);
--- a/src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
+++ b/src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace Drupal\localgov_directories\Plugin\facets\query_type;
+
+use Drupal\facets\QueryType\QueryTypePluginBase;
+use Drupal\facets\Result\Result;
+use Drupal\search_api\Query\QueryInterface;
+
+/**
+ * Provides support to "AND" facet groups while keeping the operator within a
+ * facets as an "OR".
+ *
+ * @FacetsQueryType(
+ *   id = "localgov_directories_query_type",
+ *   label = @Translation("LocalGov Directories Facet Groups AND Query Type"),
+ * )
+ */
+class LocalGovDirectoriesQueryType extends QueryTypePluginBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $query = $this->query;
+
+    // Only alter the query when there's an actual query object to alter.
+    if (!empty($query)) {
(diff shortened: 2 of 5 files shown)


finnlewis:
Thanks @snpower ! @andybroomfield and @ekes said they'd look and @ekes mentioned we probably want some tests.

andybroomfield:
Looks good from testing so far, thanks @snpower.

Before, the Just right listing should not show.

Screenshot 2022-10-04 at 1 37 22 pm

With the And filter, it is now filtered when over 18s is checked.

Screenshot 2022-10-04 at 1 38 14 pm

But returns as expected when under 18s checked.

Screenshot 2022-10-04 at 1 38 31 pm

markconroy:
@finnlewis I nabbed Stella's time over the weekend when she was offically not working! Any chance someone else from the LGD team would be able to put the necessary tests for this together; I have a feeling that might be a quicker way for us to get this complete and merged?

finnlewis:
@andybroomfield not totally clear by your screen shots - is the first shot showing that it does not work or is it all good?

finnlewis:
@Adnan-cds would you like to take a look at this too? Hoping to get it merged soon.

andybroomfield:
@finnlewis that is the before screenshot.
However the localgov and filter has to be enabled first (second screenshot).

ekes:
@andybroomfield 's looking at adding a Functional test. I've had a quick look at doing a Kernel one, but it's probably more 'prophesy' than is worth it. But while doing so I did come up with a couple of questions I'll pop in a review.

src/Plugin/facets/processor/LocalGovDirectoriesProcessor.php
```diff
@@ -0,0 +1,43 @@
...
+ *     "pre_query" = 35
+ *   }
+ * )
+ */
+class LocalGovDirectoriesProcessor extends ProcessorPluginBase implements PreQueryProcessorInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function preQuery(FacetInterface $facet) {
```
ekes, line 26: Is this needed?

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
```diff
@@ -0,0 +1,90 @@
...
+        }
+      }
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    $query_operator = $this->facet->getQueryOperator();
```
ekes, line 66: This seems like it's very similar to the Drupal\facets\Plugin\facets\query_type\SearchApiString without the missing facet error handling. Can it inherit that, rather than base, and just override the execute() that is different?

src/Plugin/facets/processor/LocalGovDirectoriesProcessor.php
snpower: sorry, is which part needed?

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
snpower: yes, it probably could alright

src/Plugin/facets/processor/LocalGovDirectoriesProcessor.php
ekes: The content of the `preQuery` just seems to getActiveItems from the $facet and then set the same value back? 
Or does this cause the facet to do something, looking it the class it doesn't seem to.

ekes (commented):
@andybroomfield Think we were agreed this should be enabled on the facet by default. So I've added it here.

finnlewis:
@andybroomfield has not had a chance to write tests for this yet, so we're not merging just yet, sorry!

finnlewis:
I might be missing the point, but I don't think this is working quite as I would expect. 

I have a few items. 

- English Government: (Country: England, Org type: Government association)
- Scottish Government: (Country: Scotland, Org type: Government association)
- Welsh Government: (Country: Wales, Org type: Government association)
- Scottish Local Authority: (Country: Scotland, Org type: Local Authority)

image

When I check England, I see only the England result, which is expected, but I can no longer see the other facets to be able to select Scotland: 

image

I think what I am trying to do is be able to select Enland and Scotland to be able see a list of items that are tagged Enland OR Scotland.

Then be able to reduce that set by selecting an Organisation Type. 

Is there a way we can get the other facets with a group to be displayed?

andybroomfield:
Possible related https://www.drupal.org/project/facets/issues/3276669

finnlewis:
@markconroy could you run this by your team and explore the UX that I describe above? Is this intended? Can we find a way to work around this specific case?

@3mkay ?

andybroomfield:
The quick and dirty method is just to re-run the query without facets, since that is what is used to generate them (means all applied facets are visible).

````php
  /**
   * {@inheritdoc}
   */
  public function build() {
    $query_operator = $this->facet->getQueryOperator();
    $temp_query = $this->query->getOriginalQuery();
    // dpm(get_class_methods($temp_query));
    $temp_query->preExecute();
    // dpm(array_keys($temp_query->getOptions()));
    // dpm($this->query->getConditionGroup()->getConditions());
    $conditions = &$temp_query->getConditionGroup()->getConditions();
    // dpm($conditions);
    foreach ($conditions as $key => $condition) {
      if ($condition instanceof \Drupal\search_api\Query\ConditionGroupInterface) {
        $tags = $condition->getTags();
        foreach($tags as $tag) {
          if (strpos($tag, 'facet:localgov_directory_facets_filter.') === 0) {
            unset($conditions[$key]);
          }
        }
      }
    }
    $temp_query->execute();
    $avalible_facets = $temp_query->getResults()->getExtraData('search_api_facets');
    $results = $avalible_facets['localgov_directory_facets_filter'] ?? $this->results;

    if (!empty($results)) {
      $facet_results = [];
      foreach ($results as $result) {
        if ($result['count'] || $query_operator === 'or') {
          $result_filter = $result['filter'] ?? '';
          if ($result_filter[0] === '"') {
            $result_filter = substr($result_filter, 1);
          }
          if ($result_filter[strlen($result_filter) - 1] === '"') {
            $result_filter = substr($result_filter, 0, -1);
          }
          $count = $result['count'];
          $result = new Result($this->facet, $result_filter, $result_filter, $count);
          $facet_results[] = $result;
        }
      }
      $this->facet->setResults($facet_results);
    }

    return $this->facet;
  }
````

markconroy:
I had a chat with @3mkay about this, and it's starting to look like 

1. the facets are working the way they should, and
2. what is being asked for is "filters" and not "facets"

@3mkay can you chip in if there's any further details needed?

3mkay:
@finnlewis there's an option in the facets settings labelled "Hide non-narrowing results", is that checked in your example? 

If yes, can you try it with that un-checked.

finnlewis:
@snpower @markconroy we're just discussing this in Tech Group Dropin and @andybroomfield is interested to try applying the code above to see if this will make the UX more intuitive. 

Should we commit to this branch or create another branch and merge request back into this one? 

We don't want to mess up anything that is already using this branch.

Cheers!

Finn

3mkay:
> Should we commit to this branch or create another branch and merge request back into this one?

A new MR based off this one would be good for us.

> We don't want to mess up anything that is already using this branch.

Thanks!

finnlewis:
From further discussion, it sounds like the behaviour I outlined above is expected. 

@ekes is happy for the motivation on this and we can either merge this in as is, or see if @andybroomfield's approach above will make it even better! 

@markconroy confirmed that there is no a huge rush for this, they already have it working.

andybroomfield:
I've started a new branch with a refined version of this code.
`feature/193-use-an-and-query-between-facet-groups-and-or-within-them-filter`

finnlewis:
Discussing at Merge Monday.

@andybroomfield is keen to know how this is playing out at Essex. @3mkay do you know who to talk to to find out how this is being used? 

Our motivation is to understand whether to go with how this pull request is working as is, or explore the branch that @andybroomfield created with the enhancements.

andybroomfield:
Just to add, I would be happy with this functionality to go out sooner and we can then work a fix on my other branch. My hunch is that real world use the issue above is not something that is relevant as much as most directory entries have good cross catogrisation accross the filters.

andybroomfield:
Here is an example of where it might be an issue, though it took a few ticks to get there.
https://send.essex.gov.uk/search-support-groups-and-activities?f%5B0%5D=localgov_directories_facets%3A3&f%5B1%5D=localgov_directories_facets%3A19&f%5B2%5D=localgov_directories_facets%3A20&f%5B3%5D=localgov_directories_facets%3A21&f%5B4%5D=localgov_directories_facets%3A27&f%5B5%5D=localgov_directories_facets%3A57

As a user, I would expect to now expand my options for areas covered, special needs and support type.

Screenshot 2022-11-15 at 2 33 00 pm

finnlewis:
We're keen to push forward with this and bring it in if it is an improvement.

@andybroomfield wants to have a look at it again. 

In practise we think this is good to go ahead with. 

@andybroomfield will test soon with some live BHCC directoires.

finnlewis:
Update from Merge Monday call, @andybroomfield is hoping to test with real live directories in the next week or two. 

Otherwise we're all happy to merge this.

Watch this space!

finnlewis:
On @andybroomfield 's backlog for the forthcoming sprint, where they'll  be looking at directoires.

finnlewis:
@andybroomfield and team in the process of testing this and working on some changes which will come back as a PR into this branch.

finnlewis:
@ekes is working on https://github.com/localgovdrupal/localgov_directories/pull/269 with @andybroomfield 

Once that is merged into this branch we should be a good to go! :)

willguv:
@finnlewis @andybroomfield can I see this working somewhere? Many thanks!

finnlewis:
Interesting question @willguv,  I wonder if we can spin up a Gitpod instance with this branch to demonstrate?

finnlewis:
@andybroomfield will get back to this at some point, but not top of the list. Mainly just needs to review comment from @ekes above.

finnlewis:
Just checking in on this at Merge Monday. @andybroomfield is hoping to be able to progress the other pull request https://github.com/localgovdrupal/localgov_directories/pull/269 and then get back to this one. 

From discussion today, we think we should re-target this to the 3.x branch to support Drupal 10 initially, then back-port to the 2.x / Drupal 9 branch if necessary. 

@snpower @markconroy any issues with re-targeting this pull request to the 3.x branch by creating a new branch and new pull request?

markconroy:
Fine with me.

andybroomfield:
Would it b e ok if I rebased this to 3.x?

markconroy:
I don't see any issue with that.

andybroomfield:
I've converted this back to draft as it needs to be rebased on 3.x.
Would there be any objection to merging in #2141 as there is progress being made on that now and we then only have one PR to review?

finnlewis:
Discussing this in Tech Drop-in, @NDITGC needs this!

@andybroomfield and @ekes in helpting to test this and https://github.com/localgovdrupal/localgov_directories/pull/235 to bring into the fold.

Note: @NDITGC is still on Drupal 9, so it would be valuable to get this merged to the Drupal 9 branches and Drupal 10.

finnlewis:
@andybroomfield and @snpower is this one now redundant as we merged https://github.com/localgovdrupal/localgov_directories/pull/325 ?

finnlewis:
Closing this - thanks for the patience!

PR #241, opened 2022-10-17
Disable facets field in the default display of the promo page

msayoung:


finnlewis (approved):
Sounds good to me, @msayoung - not tested but looks like trivial change that is sensible.

markconroy (approved):

PR #245, opened 2022-10-27
Fix: Channel search block's cache context

### LINKED ISSUE #246: Search block appearing with pre-populated keyword
## How to reproduce
1. Open a Directory channel page as an anonymous user.
2. Copy the URL from the browser's address bar.
3. Clear **Drupal cache**.
4. Search the channel from the Directory search form in the left sidebar.
5. Load the URL from step 2.
6. The search block should be **pre-populated** even though no search has been performed in the previous step.  This is due to wrong cache context usage by the Directory search block.

Bug reported by site visitors.

Adnan-cds:
Bug fix for cache context.  When the channel search block uses only the 'route' cache context, http query parameters make no difference to block caching.  To resolve this, we use the same cache context used by the [Views exposed filter block](https://git.drupalcode.org/project/drupal/-/blob/9.5.x/core/modules/views/src/Plugin/Block/ViewsExposedFilterBlock.php#L23).

### DIFF
--- a/src/Plugin/Block/ChannelSearchBlock.php
+++ b/src/Plugin/Block/ChannelSearchBlock.php
@@ -156,7 +156,10 @@ class ChannelSearchBlock extends BlockBase implements ContainerFactoryPluginInte
    */
   public function getCacheContexts() {
 
-    return Cache::mergeContexts(parent::getCacheContexts(), ['route']);
+    $channel_view = Views::getView(self::CHANNEL_VIEW);
+    $channel_view->setDisplay(self::CHANNEL_VIEW_DISPLAY);
+    $contexts = $channel_view->display_handler->getCacheMetadata()->getCacheContexts();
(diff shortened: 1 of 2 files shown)


ekes:
Ignoring the test failures because of the schema issue elsewhere.

This is similar to https://github.com/localgovdrupal/localgov_search/pull/50 right? Can we make a similar test for this https://github.com/localgovdrupal/localgov_search/pull/50/files#diff-207523791ea11407bc57be49a9e27245c0f0082f23881b556648d3d07a8f8102R157 ?

Adnan-cds:
Thanks Ekes.  Yes, I am writing a similar test right now :)

ekes (approved):
Tests pass (and fail without cache context change).
New context is like:
[
"route",
"languages:language_interface"
"url",
"url.query_args",
"user.node_grants:view",
]
which makes sense including the query_args.

Adnan-cds:
Thanks Ekes.  Merging...

PR #247, opened 2022-11-07
2.3.6 release

finnlewis:


finnlewis:
In our Merge Monday call. 

This is a bug that is annoying.

Currently are failing across the board on Functional Javascript, see https://github.com/localgovdrupal/localgov_project/actions/runs/3408292216/jobs/5668763714#step:5:22

We're going to go ahead a and release this!

No objections on the call.

:)

PR #252, opened 2022-12-06
Add a URL alias to a Directory Promo Page

msayoung:


ekes (approved):
Same as pattern for localgov_directory_page the directory is added in localgov_directories_pathauto_pattern_alter

PR #253, opened 2022-12-12
2.4.0 release

ekes:


ekes:
MM this should include a update hook to add in the path alias if installed and none yet set.

ekes:
https://github.com/localgovdrupal/localgov_directories/pull/255 for the update hook.

finnlewis:
waiting for https://github.com/localgovdrupal/localgov_directories/pull/255

finnlewis:
This release now includes potentail API changes 

https://github.com/localgovdrupal/localgov_directories/pull/253/files#diff-118751fdb44d76d8d3f435d224b55591883bf325ae2fd136581f092d0f28534aL127

So @ekes suggests that strictly speaking, we should bump the minor version number to 2.4.0

We can review on Monday but any comments welcome before @andybroomfield @Adnan-cds @stephen-cox

andybroomfield:
2.4.0 would be safer and technically correct if there is an api break risk.
I would be tempted to also mark them as internal if thats intended, though that can wait till 2.4.1 I suppose.

finnlewis (approved):
Looks good, we'll make release notes that explain.

Issue #254, opened 2022-12-12
Write update hook for promo page path auto


ekes:
Add pathauto config if it does not yet exist to sites with promo page already enabled.

For new installs added here https://github.com/localgovdrupal/localgov_directories/pull/252

ekes:
https://github.com/localgovdrupal/localgov_directories/pull/255

PR #255, opened 2022-12-12
Add update hook to add pathauto pattern should it not exist yet #254.

ekes:
To test:
 * Install version of localgov_directory_promo_page _before_ https://github.com/localgovdrupal/localgov_directories/commit/e142ff22438ed025d511c71f4fc12fa451e2e789 so current 2.3.7 version will do.
 * Enable localgov_directory_promo_page module
 * Then switch to this branch and run updb

The pathauto pattern should be exactly the same, as installing from scratch at HEAD.

Stretch goal. Do the same with a pathauto pattern added, it shouldn't change.

### DIFF
--- a/modules/localgov_directories_promo_page/localgov_directories_promo_page.install
+++ b/modules/localgov_directories_promo_page/localgov_directories_promo_page.install
@@ -5,13 +5,18 @@
  * Update functions for the LocalGov Directories Promo Page.
  */
 
+use Drupal\node\Entity\Node;
 use Drupal\pathauto\Entity\PathautoPattern;
 
 /**
  * Adds pathauto pattern if there is not one yet.
  */
 function localgov_directories_promo_page_update_8001() {
-  $pattern = PathautoPattern::load('localgov_directory_promo_page');
+  $pathauto = \Drupal::service('pathauto.generator');
+  $entity = Node::create([
+    'type' => 'localgov_directory_promo_page',
+  ]);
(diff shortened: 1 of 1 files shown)


andybroomfield (approved):
This worked for me, testing the scenario whereby installing without the pattern and them running drush updb now adds the correct pattern.

I havn't tested the scenario whereby a pattern had been already set up, as this would only test the machine name anyway both patterns may end up present and site builders would have to sort out which one to keep.

finnlewis:
Discussing at Merge Monday. @andybroomfield pointed out that this loads the url alias pattern based on machine name. 

If someone has already created a path auto pattern manually, this will currently introduce a duplicate. 

We could cover this in release notes, but probably better to try to load a path auto pattern that matches on content type = directory promotional page.

@ekes thinks this is possible - just needs looking at.

ekes:
Manual testing pattern:-

Install (without this branch). Then
a) just switch to this branch and upgrade you should get the pathauto pattern;
Repeat but this time
b) create a path auto pattern machine name anything you like for the promo page; then switch to this branch and upgrade; you shouldn't get the pathauto pattern.

finnlewis:
Thanks, I will test

finnlewis (approved):
Yes mate, this works a treat! Confirmed that it adds the pathauto pattern if one does not exists and a pattern is already there for the localgov_directory_promo_page, it leaves it well alone.

Nice one @ekes ! I'll merge this and prep the release for review on Monday as discussed.

PR #266, opened 2023-02-15
loads maps correctly when a tab is clicked

### LINKED ISSUE #265: If an geofield map is in a vertical (or other) tab, then it doesn't load properly unless you resize the screen.
This is what we currently have:

Screenshot 2023-02-15 at 16 51 11

This is what we want to have:
Screenshot 2023-02-15 at 16 51 25

markconroy:
Closes #265

### DIFF
--- a/js/localgov_directories_maps.js
+++ b/js/localgov_directories_maps.js
@@ -5,7 +5,6 @@ Drupal.behaviors.localgovDirectoriesMaps = {
 
     maps.forEach(map => {
       const mapParentId = map.closest('.field-group-tab').getAttribute('id');
-      console.log(mapParentId);
       const correspondingLink = context.querySelector(`[href="#${mapParentId}"]`);
       correspondingLink.addEventListener('click', () => {
         window.dispatchEvent(new Event('resize'));
--- a/js/localgov_directories_maps.js
+++ b/js/localgov_directories_maps.js
@@ -1,0 +1,16 @@
+Drupal.behaviors.localgovDirectoriesMaps = {
+  attach: function (context, settings) {
+    context = context || document;
+    const maps = context.querySelectorAll('form .geofield-map-widget');
+
+    maps.forEach(map => {
+      const mapParentId = map.closest('.field-group-tab').getAttribute('id');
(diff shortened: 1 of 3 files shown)


ekes:
This is pretty specific to only one case in https://github.com/localgovdrupal/localgov_directories/issues/265#issuecomment-1433174061 I think there are actually more.

https://github.com/localgovdrupal/localgov_directories/issues/265#issuecomment-1433331027 will do it for pretty much any case where the map is not rendered in the initial viewport, so hopefully covering all the cases, just a question of how far up the module chain we put it our geo entity, geofield map or leaflet.

ekes:
Looks like [itamair](https://www.drupal.org/u/itamair) is up for putting this directly into geofield_map and leaflet https://www.drupal.org/project/geofield_map/issues/3087072#comment-14930191
If not, or if we need it sooner, I think it should be on localgovdrupal/localgov_geo

finnlewis:
Note: @markconroy has a custom module that does whay he needs for now, so happy to wait to see if we can get this in upstream in lealeft and geofied_map modules on drupal.org.

ekes:
So the patch for leaflet went in, but not for geofield_map. Now... we don't actually need the geofield map widget anymore. I've written stuff to do the geocoding into the leaflet map widget. It's in geo_entity at the moment, but should be ported into localgov_geo. Then we can just drop geofield_map.

finnlewis:
Discussing in Merge Monday: this pull request will probably go away once we add a patch from: https://git.drupalcode.org/issue/geofield_map-3087072/-/compare/8.x-2.x...3087072-visibility-issue-with

See also: https://www.drupal.org/project/leaflet/issues/3343154

But then we want to stop using geofield_map (@ekes has done the work for this but ,might need backporting from geofield_entity)

finnlewis:
Note to self: check on status of patch inclusion above.

markconroy:
I think we can close this PR, since it's being worked on in a Drupal.org issue instead. @ekes are you okay with us closing this?

finnlewis:
@ekes shall we close this?

PR #269, opened 2023-03-17
Try to find all the possible sibling facets from a facet group

andybroomfield:
To try and resolve this comment 
  https://github.com/localgovdrupal/localgov_directories/pull/235#issuecomment-1277842772

Adds a filter for the facets to try and get the sibling versions so possible OR facets are avalible if the other groups set an AND.

This works well for two facets groups, though more does lead to more 0 result possiblilties.

### DIFF
--- a/src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
+++ b/src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
@@ -71,7 +71,7 @@ class LocalGovDirectoriesQueryType extends QueryTypePluginBase {
     // Get passed in facets.
     $temp_query = $this->query->getOriginalQuery();
     $temp_query->preExecute();
-    $conditions = &$temp_query->getConditionGroup()->getConditions();
+    $conditions = $temp_query->getConditionGroup()->getConditions();
 
     // Find all the facet condition groups.
     $facet_conditions = [];
@@ -147,7 +147,7 @@ class LocalGovDirectoriesQueryType extends QueryTypePluginBase {
     $filter_query->preExecute();
 
     // Find conditions.
-    $conditions = &$filter_query->getConditionGroup()->getConditions();
+    $conditions = $filter_query->getConditionGroup()->getConditions();
 
     // Store removed conditions so we can reset them.
     $removed_conditions = [];
@@ -172,10 +172,8 @@ class LocalGovDirectoriesQueryType extends QueryTypePluginBase {
     $facets = $filter_query->getResults()->getExtraData('search_api_facets');
     $filter_query->postExecute();
 
-    // Beacuse for reasons unknown, removing the conditions removes it from all
-    // the following queries, even though we are fetching the original query.
-    // We can get around that by readding the conditions back in now that we
-    // have the facets we want.
+    // Without deep cloning we've affected the conditions, reset these for the
+    // query.
     $conditions = array_merge($conditions, $removed_conditions);
 
     // Since we will get every facet from the passed in facet group,
--- a/src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
+++ b/src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
@@ -66,32 +66,43 @@ class LocalGovDirectoriesQueryType extends QueryTypePluginBase {
   public function build() {
     $results = $this->results;
     $query_operator = $this->facet->getQueryOperator();
+    $field_identifier = $this->facet->getFieldIdentifier();
+
+    // Get passed in facets.
     $temp_query = $this->query->getOriginalQuery();
     $temp_query->preExecute();
     $conditions = &$temp_query->getConditionGroup()->getConditions();
+
+    // Find all the facet condition groups.
+    $facet_conditions = [];
     foreach ($conditions as $key => $condition) {
-      if ($condition instanceof \Drupal\search_api\Query\ConditionGroupInterface) {
+      if ($condition instanceof ConditionGroupInterface) {
         $tags = $condition->getTags();
         foreach($tags as $tag) {
-          if (strpos($tag, 'facet:localgov_directory_facets_filter.') === 0) {
-            // unset($conditions[$key]);
-            $results = array_merge($results, $this->getGroupFacets($tag));
+          if (strpos($tag, 'facet:' . $field_identifier . '.') === 0) {
+            $facet_conditions[$key] = $tag;
           }
         }
       }
     }
-    $filtered_results[] = reset($results);
-    foreach ($results as $result) {
-      $current_results = array_column($filtered_results, 'filter');
-      if (!in_array($result['filter'], $current_results)) {
-        $filtered_results[] = $result;
-      }
-    }
-    $results = $filtered_results;
-    // $temp_query->execute();
-    // $avalible_facets = $temp_query->getResults()->getExtraData('search_api_facets');
-    // $results = $avalible_facets['localgov_directory_facets_filter'] ?? $this->results;
 
+    // Run query elimnating each facet group and return the resulting facets.
+    $results = $this->results;
+    foreach($facet_conditions as $key => $filter_tag) {
+      $group_facets = $this->getResultingFacetsFromFacetGroupExceptOwn($filter_tag);
+      
+      // Remove any duplicate facets, or they show up multiple times.
+      $group_facets_filtered = array_filter($group_facets, function($item) use ($results) {
+        $facet_ids = array_column($results, 'filter');
+        if (in_array($item['filter'], $facet_ids)) {
+          return FALSE;
+        }
+        return TRUE;
+      });
+      $results = array_merge($results, $group_facets_filtered);
+    }
+
+    // Build facets.
     if (!empty($results)) {
       $facet_results = [];
       foreach ($results as $result) {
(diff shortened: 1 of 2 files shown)


src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
```diff
@@ -87,4 +125,78 @@ public function build() {
...
+    // Store removed conditions so we can reset them.
+    $removed_conditions = [];
+
+    // Loop through each conditions, removing ones that are not this filter tag.
+    foreach ($conditions as $tag => $condition) {
+      if ($condition instanceof ConditionGroupInterface) {
+        $tags = $condition->getTags();
+
+        // @todo Check that we are only removing facet conditions.
+        if (!in_array($filter_tag, $tags)) {
```
andybroomfield: We could also reverse this proccess, so that we elimnate a specific facet group and find all the other facets that will fit, its a bit of a trial and error and could also lead to the active ones being hidden.

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
```diff
@@ -87,4 +125,78 @@ public function build() {
...
+    // show us only the ones that are reachable from this facet group.
+    $facet_type_id = substr($filter_tag, 39);
+    $group_facet_ids = \Drupal::entityTypeManager()
+      ->getStorage('localgov_directories_facets')
+      ->getQuery()
+      ->condition('bundle', $facet_type_id)
+      ->execute();
+    $found_facets = $facets['localgov_directory_facets_filter'] ?? [];
+    $found_facets = array_filter($found_facets, function($item) use($group_facet_ids) {
+      if (!in_array(intval(trim($item['filter'], '"')), $group_facet_ids)) {
```
andybroomfield: If we do decide to reverse the facet filtering, then only filter this groups facets instead of here where we filter every other group.

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
```diff
@@ -63,11 +64,48 @@ public function execute() {
...
+      
+      // Remove any duplicate facets, or they show up multiple times.
+      $group_facets_filtered = array_filter($group_facets, function($item) use ($results) {
+        $facet_ids = array_column($results, 'filter');
+        if (in_array($item['filter'], $facet_ids)) {
+          return FALSE;
+        }
+        return TRUE;
+      });
+      $results = array_merge($results, $group_facets_filtered);
```
andybroomfield, line 101: Theres still a risk that the active chosen facets get filtered, and theres no way of unticking them, so we need a way to keep in the existing facets.

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
```diff
@@ -87,4 +125,78 @@ public function build() {
...
+      }
+    }
+
+    // Execute the filter query and get the facets returned.
+    $filter_query->execute();
+    $facets = $filter_query->getResults()->getExtraData('search_api_facets');
+    $filter_query->postExecute();
+
+    // Beacuse for reasons unknown, removing the conditions removes it from all
+    // the following queries, even though we are fetching the original query.
```
ekes: This is because clone only clones the entity itself, references to other entities themselves as references. So it's still a reference to the Condition entity. Thus directly changing the referenced (shared) entity.

One option could be to use DeepCopy that's included in Drupal because (I think of simpletest) but as we don't actually need a fully deep clone can we rather just also clone the Condition and swap that clone on the cloned query?

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
```diff
@@ -63,11 +64,48 @@ public function execute() {
...
    */
   public function build() {
+    $results = $this->results;
     $query_operator = $this->facet->getQueryOperator();
+    $field_identifier = $this->facet->getFieldIdentifier();
 
-    if (!empty($this->results)) {
+    // Get passed in facets.
+    $temp_query = $this->query->getOriginalQuery();
+    $temp_query->preExecute();
```
ekes, line 73: What's the reason for getting the original query and running pre execute here? Maybe related cloning issue later. Can the actual query be used?

I noticed it because by doing so it's running execute() of this LocalGovDirectoriesQueryType again which was all getting a bit circular.

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
```diff
@@ -63,11 +64,48 @@ public function execute() {
...
+    $conditions = &$temp_query->getConditionGroup()->getConditions();
```
ekes: So yes this is by reference already as it's an object.

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
```diff
@@ -63,11 +64,48 @@ public function execute() {
...
+        foreach($tags as $tag) {
+          if (strpos($tag, 'facet:' . $field_identifier . '.') === 0) {
+            $facet_conditions[$key] = $tag;
+          }
+        }
+      }
+    }
+
+    // Run query elimnating each facet group and return the resulting facets.
+    $results = $this->results;
```
ekes: Duplicates https://github.com/localgovdrupal/localgov_directories/pull/269/files#diff-0872c6a42a990750a6b255b18dd8466e3dc6af43670c3755fae3c1e155fca46dR67 and we've not used it yet?

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
ekes: 🤯 still not quite grok'd this. Is there something we could test maybe? Or does it matter enough?

ekes:
As we're preExcute'ing the query a couple of times in here, we're calling our own ::execute() method a couple of times more too. So while I get that we need to make multiple queries to ascertain which facets to remove, I wonder if ::build is the place to do it. However I can't say I have quite a better solution off the top of my head. And it seems to be working, even if multiplying the queries later on.

finnlewis:
@andybroomfield putting this back on the list :)

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
andybroomfield: With preexecute
Screenshot 2023-05-24 at 6 27 12 pm


Without preexecute
Screenshot 2023-05-24 at 6 27 26 pm

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
andybroomfield: @ekes Would you be able to provide a hint on how to clone just the conditions group for the filter query?
Wouldn't we still need to re-add them though if everything is a clone? I'm not sure how we can actully avoid doing the current method in a way that is not overrly complex.

finnlewis:
@ekes will take another look to review the logic of cloning conditions....  @andybroomfield was a bit lost on the proposed approach

ekes:
[5c32a6a](https://github.com/localgovdrupal/localgov_directories/pull/269/commits/5c32a6a71ec797932f94e09ad72269c36ad1d0f9) keeps the reusing the query approach, avoids the deep cloning that @andybroomfield was mentioning, but makes explicit why and when it's doing this.

andybroomfield:
Thanks @ekes will take a review of this and see if we can move this forward.

finnlewis:
Other things have taken @andybroomfield 's time.... 

We might want to rebase this against the Drupal 10 branch (3.x)

The need for AND operator on facets / filters is currents (North Devon) so we should get this rebased, reviewed and released soon. 

Then we need to get back to https://github.com/localgovdrupal/localgov_directories/pull/235 where the merge request has been open for over a year!

finnlewis:
@NDITGC CC

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
```diff
@@ -147,7 +147,7 @@ protected function getResultingFacetsFromFacetGroupExceptOwn(string $filter_tag)
     $filter_query->preExecute();
 
     // Find conditions.
-    $conditions = &$filter_query->getConditionGroup()->getConditions();
+    $conditions = $filter_query->getConditionGroup()->getConditions();
```
andybroomfield: My finding is we need the `&` and take this by reference or we get strange behaviour where the conditions don't get reset.
To test, I added countries to the demo Money, employment and training directory.

Without taking the query by reference.
Screenshot 2023-10-16 at 4 59 19 pm

With query by reference
Screenshot 2023-10-17 at 10 37 25 am

andybroomfield: Note: Northern ireland should still be visible even when not ticked.

Screenshot 2023-10-16 at 4 59 37 pm

andybroomfield:
Going back to draft as I need to rebase this and the parent to 3.x

finnlewis:
@andybroomfield is this still live?

andybroomfield:
Unless you want to support Drupal 9 this and the related PR can be closed...

andybroomfield:
I'm going to close this as this was the Drupal 9 branch and I think most people are focused on upgrading to Drupal 10 which does have this facet plugin.

PR #270, opened 2023-03-20
Remove non-active facets from facet display on channel page.

ekes:


finnlewis:
Need to take this for a test drive.

ekes:
pre-MM ping

ekes:
@finnlewis post-MM ping

finnlewis:
@ekes will take a look at the failing test. 

If anyone else can do a manual test, that would be great ! @Adnan-cds @stephen-cox @andybroomfield

src/DirectoryExtraFieldDisplay.php
```diff
@@ -286,6 +298,19 @@ public function preprocessFacetList(array &$variables) {
...
+
+    // This is usually on a channel node. If so remove facets not active on
+    // channel.
+    if (($channel = $this->routeMatch->getParameter('node'))
+      && $channel instanceof NodeInterface
+      && $channel->bundle() == 'localgov_directory'
+    ) {
+      $active_facets = array_column($channel->localgov_directory_facets_enable->getValue(), 'target_id');
+    }
+    if ($active_facets) {
```
Adnan-cds: Can this lead to a PHP warning if we are not on a localgov_directory node?

Adnan-cds:
> If anyone else can do a manual test, that would be great !

Tried this on a test site with a small dataset and it worked as expected :)

src/DirectoryExtraFieldDisplay.php
ekes: Yep. That's one of the test fails. Should be updated with [31db4e5](https://github.com/localgovdrupal/localgov_directories/pull/270/commits/31db4e5a6f16dbf04f653926c29766ae8b0459e5)

src/DirectoryExtraFieldDisplay.php
```diff
@@ -291,6 +303,20 @@ public function preprocessFacetList(array &$variables) {
...
+    // This is usually on a channel node. If so remove facets not active on
+    // channel.
+    $active_facets = NULL;
+    if (($channel = $this->routeMatch->getParameter('node'))
+      && $channel instanceof NodeInterface
+      && $channel->bundle() == 'localgov_directory'
+    ) {
+      $active_facets = array_column($channel->localgov_directory_facets_enable->getValue(), 'target_id');
+    }
+    if (!is_null($active_facets)) {
```
Adnan-cds, line 316: I am okay with this.  Dropping line 309 and changing 316 to `if (!empty($active_facets)) {` would also do.  Sorry, should have mentioned this earlier :(

Adnan-cds (approved):
Works for me.  Thanks.

src/DirectoryExtraFieldDisplay.php
ekes: After reading a blog post, I can't remember where, I'm quite liking avoiding using `empty` and trying to be more explicit. Which in this case means I guess we understand array_column could technically return an empty array and this get passed into array_intersect_key.

Issue #276, opened 2023-03-23
Make adding 'facet types' more straightfoward


willguv:
Clicking on "directory facets" in the content sub menu...

![Image](https://user-images.githubusercontent.com/708140/227202985-672207cc-487e-4594-85cb-106b799e96ef.png)

... takes users to this page

![Image](https://user-images.githubusercontent.com/708140/227203046-def88f8b-2c3f-4721-93af-1ac35962de4c.png)

- Facets can be added here
- Facet types cannot

To add a facet type, users need to go to the top left content menu and navigate from there

![Image](https://user-images.githubusercontent.com/708140/227203073-78f928a5-1803-4d24-be4b-18199bcba6f0.png)

How can this be made more straightforward?

willguv:
@stephen-cox this feels like a trivial fix - is it?

Many thanks!

stephen-cox:
@willguv It's easy to add a 'Add directory facet type' button to the top of the page next to the 'Add directory facet' button.

It might be worth overhauling the UX of managing directory facets entirely, so they're managed more like taxonomy terms. There would be a page that lists all the types, clicking on one would take you to another page where all the facets for that type can be managed. This approach isn't really compatible with having directory facets managed through the content overview pages though as these are views of content.

That said, we can do both. Keep the content page and add a button to add new types while creating a more intuitive way to manage directory facets.

willguv:
@stephen-cox if it's quick, can you add the 'add directory facet type' button and I'll write another card to consider the UX of directory facets. Thanks again

stephen-cox:
@willguv PR #277 adds this button so the directory facet management page looks like

![image](https://user-images.githubusercontent.com/7189914/230363532-45ecb895-9bb7-4f29-966f-626ad82f76ef.png)

The one issue I see with this is that once you've added a new directory facet type you are redirected to the directory facet type management page (/admin/structure/localgov_directories_facets_types) and so to start adding new facets of this type you'll need to go back to the directory facet management page.

I'm not sure there's an easy fix for this.

willguv:
Thanks @stephen-cox. I'm away from my computer for a couple of days so will have a look when I return

andybroomfield:
Maybe add a ?destination= query string to the add facet type link?

stephen-cox:
> Maybe add a ?destination= query string to the add facet type link?

[...]

PR #277, opened 2023-04-06
Add 'Add directory facet type' button to directory facets management page

stephen-cox:
For discussion see: https://github.com/localgovdrupal/localgov_directories/issues/276#issuecomment-1498916345

### DIFF
--- a/localgov_directories.links.action.yml
+++ b/localgov_directories.links.action.yml
@@ -3,7 +3,14 @@ localgov_directories_facets.type_add:
   title: 'Add directory facets type'
   appears_on:
     - entity.localgov_directories_facets_type.collection
+localgov_directories_facets.type_add_from_facets:
+  route_name: entity.localgov_directories_facets_type.add_form
+  title: 'Add directory facets type'
+  appears_on:
     - entity.localgov_directories_facets.collection
+  options:
+    query:
+      destination: '/admin/content/directories/facets'
 localgov_directories_facets.add_page:
   route_name: entity.localgov_directories_facets.add_page
   title: 'Add directory facets'
--- a/localgov_directories.links.action.yml
+++ b/localgov_directories.links.action.yml
@@ -3,6 +3,7 @@ localgov_directories_facets.type_add:
   title: 'Add directory facets type'
   appears_on:
     - entity.localgov_directories_facets_type.collection
+    - entity.localgov_directories_facets.collection
 localgov_directories_facets.add_page:
   route_name: entity.localgov_directories_facets.add_page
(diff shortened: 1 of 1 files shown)


finnlewis:
Not ready just yet, need to check the UX and workflow of adding facet types and facets.

willguv:
@stephen-cox could I see this working somewhere please? Many thanks

stephen-cox:
> @stephen-cox could I see this working somewhere please? Many thanks

@willguv I've deployed it here: https://test.localgovdrupal.org/admin/content/directories/facets

willguv:
Thanks @stephen-cox agree it's not perfect, but it's an improvement on what we have now. Users can add facet types from /admin/content/directories/facets which they were unable to do before

Would it be a lot of work to direct users adding a facet type from https://test.localgovdrupal.org/admin/content/directories/facets back to that page once added? If they then want to add a facet to that type, there's a clear route

Let's talk separately about an overhaul

stephen-cox:
We would need to implement some sort of hook to add the redirect, it shouldn't be too much work, but would need to investigate to determine the best way to implement this.

One issue that the redirect would apply to anyone who added a directory facet, which is not necessarily the best behavior; someone might want to add a bunch of facet types before then adding the facets. We could check the referrer header and only redirect when the user came from the facet management page, but this feels a bit hacky.

willguv:
@stephen-cox could you describe how to rework this in the taxonomy style? Am trying to picture it - thanks

finnlewis:
@stephen-cox I've just added a new action link with a destination to return users to the facet content page after adding a facet from there. 

So now, if we are at admin/content/directories/facets and click on "Add directory facet type" we can add a facet type and get returned to admin/content/directories/facets

If we are admin/structure/localgov_directories_facets_types and click on "Add directory facet type" we can add a facet type and get returned to admin/structure/localgov_directories_facets_types

Does that work?

stephen-cox:
> @stephen-cox could you describe how to rework this in the taxonomy style? Am trying to picture it - thanks

@willguv What I had i mind is what's been implemented for Microsites

The the main directory facets page lists the types:

image

Clicking on a type takes you to the facets for that type

image

willguv:
@stephen-cox that's much clearer - how long to implement please?

Will there be any issues for existing installs?

stephen-cox:
> @stephen-cox I've just added a new action link with a destination to return users to the facet content page after adding a facet from there.
> Does that work?

Nice one @finnlewis Does the job :+1:

ekes (approved):

PR #278, opened 2023-04-17
Fix #275: Only import services navigation field optional configuration.

### LINKED ISSUE #275: Only install localgov_directories optional config when modules are installed
BHCC (and I think others) noticed that an update to 2.5.0 enabled a bunch of optional config in all modules
(Notably the core node book content type and some core image styles we deleted).

See line 80 - 89 in localgov_directories.module

````php
/**
 * Implements hook_modules_installed().
 */
function localgov_directories_modules_installed($modules) {
  $services = in_array('localgov_services_navigation', $modules);
  if ($services) {
    \Drupal::service('config.installer')->installOptionalConfig();
    localgov_directories_optional_fields_settings($services);
  }
}
````

Am I right in thinking that this will install all optional config if the localgov_services_navigation is one of the modules installed.
(I'm assuming this is here so that if localgov_services is enabled after directories, the parent field gets added to the channels.)
However, reading [public function ConfigInstallerInterface::installOptionalConfig](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Config%21ConfigInstallerInterface.php/function/ConfigInstallerInterface%3A%3AinstallOptionalConfig/8.2.x) there are two optional parameters that can be passed (the storage and a dependancy array) that without which all optional config from all modules will be searched through and installed if it can.

As I think the intention here is just to install this modules optional config, we should find a way to limit that.

ekes:
@andybroomfield You want to review this?
@stephen-cox As it's only mentioning navigation here I'm assuming there's no other optional field we're accidentally relying on getting imported? I've looked through the submodules optional and can't see anything.

### DIFF
--- a/localgov_directories.module
+++ b/localgov_directories.module
@@ -83,7 +83,9 @@ function localgov_directories_localgov_roles_default() {
 function localgov_directories_modules_installed($modules) {
   $services = in_array('localgov_services_navigation', $modules);
   if ($services) {
-    \Drupal::service('config.installer')->installOptionalConfig();
+    \Drupal::service('config.installer')->installOptionalConfig(NULL, [
+      'config' => 'field.storage.node.localgov_services_parent',
+    ]);
     localgov_directories_optional_fields_settings($services);
   }
 }
--- a/src/ConfigurationHelper.php
+++ b/src/ConfigurationHelper.php
@@ -324,7 +324,9 @@ class ConfigurationHelper implements ContainerInjectionInterface {
 
     $conditional_config_path = $this->moduleExtensionList->getPath('localgov_directories') . '/config/conditional';
     if ($this->importConfigEntity('facets_facet', $conditional_config_path, $facet_cfg_file)) {
-      $this->configInstaller->installOptionalConfig();
+      $this->configInstaller->installOptionalConfig(NULL, [
+        'config' => 'facets.facet.localgov_directories_facets',
+      ]);
     }
   }
 


ekes:
If this is good we should follow-up on https://github.com/localgovdrupal/localgov/issues/376 and do the same for any other places we import the optional config.

andybroomfield:
This is still exporting all optional config (re-creating the book node and old media styles).

Screenshot 2023-04-17 at 2 02 30 pm

finnlewis:
How to test this: 

- install Drupal stock (not localgov profile)
- enable directories
- confirm that the localgov_services_parent field is not here 
- enable services landing page
- see field appear

Note: once this works for a standard profile install, we can perhaps follow suit across the other instances of optional config.

ekes:
You also want to install another module and delete some optional config out of it, before enabling localgov_services_landing, just to check that doesn't get reimported again.

The reason to do this with only the few modules enabled is we implement the hook in other localgov_* modules in a similar way, so they will still be causing other optional config to be imported. The reason not to do all of them at the same time is, if this one works, when we implement it elsewhere we need to check that we're not actually relying on the others to import some config that we're not _explicitly_ mentioning yet.

andybroomfield:
After doing some further looking it is the call to `$this->configInstaller->installOptionalConfig();`
in localgov_directories/src/ConfigurationHelper.php that causes it when the facets are being resaved.
That makes sense as no new modules are being installed.
````php
  /**
   * Import config entity for the directory Facet.
   */
  public function createFacet(string $facet_id, string $facet_cfg_file): void {
    if ($this->entityTypeManager->getStorage('facets_facet')->load($facet_id)) {
      return;
    }

    $conditional_config_path = $this->moduleExtensionList->getPath('localgov_directories') . '/config/conditional';
    if ($this->importConfigEntity('facets_facet', $conditional_config_path, $facet_cfg_file)) {
      $this->configInstaller->installOptionalConfig();
    }
  }
````

Commenting out $this->configInstaller->installOptionalConfig(); does not cause the extra optional config to import, but then I guess there is a need to get optional config here?

ekes:
> After doing some further looking it is the call to `$this->configInstaller->installOptionalConfig();` in localgov_directories/src/ConfigurationHelper.php that causes it when the facets are being resaved. That makes sense as no new modules are being installed.
> 
> ```
>   /**
>    * Import config entity for the directory Facet.
>    */
>   public function createFacet(string $facet_id, string $facet_cfg_file): void {
>     if ($this->entityTypeManager->getStorage('facets_facet')->load($facet_id)) {
>       return;
>     }
> 
>     $conditional_config_path = $this->moduleExtensionList->getPath('localgov_directories') . '/config/conditional';
>     if ($this->importConfigEntity('facets_facet', $conditional_config_path, $facet_cfg_file)) {
>       $this->configInstaller->installOptionalConfig();
>     }
>   }
> ```
> 
> Commenting out $this->configInstaller->installOptionalConfig(); does not cause the extra optional config to import, but then I guess there is a need to get optional config here?

Most important thing I think is it auto imports the blocks out of the optional config (if you are using the localgov theme - ach more dependencies to test, or I suspect we actually have automated tests for it already). Anyway I've added that conditional configuration name to the installOptionalConfig there too https://github.com/localgovdrupal/localgov_directories/pull/278/commits/b1339e212e52d789db26d9bb8909d4639d65c009

ekes:
> You also want to install another module and delete some optional config out of it, before enabling localgov_services_landing, just to check that doesn't get reimported again.
> 
> The reason to do this with only the few modules enabled is we implement the hook in other localgov_* modules in a similar way, so they will still be causing other optional config to be imported. The reason not to do all of them at the same time is, if this one works, when we implement it elsewhere we need to check that we're not actually relying on the others to import some config that we're not _explicitly_ mentioning yet.

For doing it with localgov profile the other modules that shouldn't be installed are guides, step by step and events https://github.com/localgovdrupal/localgov/issues/376#issuecomment-1511544653

andybroomfield:
Looks like it is exporting correct config now (nothing being resurrected).

Screenshot 2023-04-18 at 8 36 10 pm

andybroomfield (approved):

Issue #281, opened 2023-04-24
Add documentation / description for facets


ekes:
Explain that you can use standard search api facets if you want to make them by developer and deploy; you can use directory facets if the directory needs to be configured on production by a content editor.

PR #286, opened 2023-05-15
Add additional clearer documentation about facets #281

ekes:


finnlewis (approved):
Makes sense to me :) Also covered on the docs site.

PR #290, opened 2023-06-22
3.0.0-alpha3

ekes:
Update module dependencies for D10.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,7 +86,6 @@ jobs:
 
       - name: DON'T MERGE THIS - Require Drupal 10 compatible branches
         run: |
-          composer --working-dir=./html require localgovdrupal/localgov_core:"dev-feature/drupal-10 as 2.2.0"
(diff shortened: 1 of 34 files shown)


finnlewis:
Merge Monday discussion: we should probably fix the deprecated code issues. 
(But note that the 3.x branch of localgov is currently pointing at dev branch for localgov_directories.)

stephen-cox:
I've started fixing things on #294

stephen-cox:
Still one failing test.

Fails because the `'settings[handler_settings][target_bundles][localgov_directory]'` field doesn't exist in the page with: 

https://github.com/localgovdrupal/localgov_directories/blob/3.x/tests/src/Functional/ChannelFacetsWidgetAdminTest.php#L122-L131

It exists as a hidden field on my local, but is missing in the test. Been bashing my head against it, but no luck so far

finnlewis:
Note: we already have an alpha2, this might want to be alpha3

ekes:
I've just pushed the fix for the test fails into https://github.com/localgovdrupal/localgov_directories/commit/e9dbc2ea3f82222b39b5217e2f9f546d1e509672 can be cherry picked to here if we want to get this one out first.

PR #291, opened 2023-06-26
Make facets on a channel optional, as is documented from #281.

ekes:
Doing dev on a channel that doesn't use content facet, but taxonomy facets instead, and keep getting reminded this field isn't actually compulsory. So figured rather than unticking it every time I reinstall I should make a PR :)


ekes:
(should actually be applied to 2.x and 3.x if merged)

finnlewis (approved):
Looks trivial enough to me!

PR #296, opened 2023-07-07
Re-run core update for our blocks added after the fact.

ekes:
As per comment in the code. We've added config with deprecated plugin after the core update. So this just runs it again. Doing so in all the modules we added this config would make sense.

This does to existing sites what https://github.com/localgovdrupal/localgov_directories/commit/8d9c8919d8ec4994077475286a9da8068df3f354 did for new installs.

### DIFF
--- a/localgov_directories.post_update.php
+++ b/localgov_directories.post_update.php
@@ -1,0 +1,28 @@
+<?php
+
+/**
+ * @file
+ * Post update hooks for LocalGov Directories.
+ */
+
+/**
+ * Updates the node type visibility condition.
+ *
+ * This was inluded in core D9 as block_post_update_replace_node_type_condition
+ * but we had installed condition plugins with this after it might have run.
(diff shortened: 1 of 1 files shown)


stephen-cox:
This is needed for the 3.0.0-alpha1 release of the profile: https://github.com/localgovdrupal/localgov/pull/588

stephen-cox (approved):
:+1:

finnlewis:
Merging in Ekes's absence,  to help with fixing tests.

markconroy:
Looks like if you have a block that was removed, then this update is putting that block back in place again.

For example, on a directory channels if you don't want the map, so removed that block, then run this update the map comes back in again - just had a report of this happening on a client site.

I wonder could we make the code a little more secure to only update/reset blocks that have not be removed/edited? Or at least see if we could make this a consideration if we need to run a similar update on our other modules.

PR #299, opened 2023-07-26
Fixes the missing show reset facet link

MattOz-CDS:
When the configuration option to add a facet reset link is chosen, it is added to the facet array even though it is not technically facet.


stephen-cox:
@ekes is going to take a look at this

ekes (changes requested):
@MattOz-CDS nice one.

I've manually applied the changes for the fix (ie the diff on `src/DirectoryExtraFieldDisplay.php` and the new file `tests/src/Functional/ResetFacetFilterTest.php`) and it looks good.

However because it's mixed in with the changes to the use statement ordering it's a bit difficult to review, and now it doesn't merge because some (or all?) of them have been corrected in the main `2.x` branch.

Would it be possible to make a new PR with _just_ the two changes I mention that are needed for the fix to the facet? And if there are still use statements in the wrong order do make a second separate one to reorder them. But keeping them separate will make it easier to review, and stop this merge conflict issues.

Thanks for all your work on this!

MattOz-CDS:
Thanks for taking the time to review this @ekes . I've applied your recommendations to [a new branch](https://github.com/localgovdrupal/localgov_directories/pull/305)

PR #305, opened 2023-08-09
[2.x] Fixes the missing show reset link (2)

MattOz-CDS:
I recreated the branch based

PR #308, opened 2023-08-25
[3.x] Fixes the missing show reset link 

ekes:
D10 fix for showing the 'reset all' facet option. D9 version #305


ekes:
This doesn't need another thorough review I don't think. I've already done it. Just tagged reviewers so some gives it a sanity scan before merging (on monday?) along with #305

PR #314, opened 2023-09-22
Fix: Suppress PHP 8.1 and 8.2 deprecation warnings

Adnan-cds:
Callable names like `static::foo` are now deprecated.  So are use of undeclared dynamic object properties.

@see https://www.php.net/releases/8.1/en.php#first_class_callable_syntax
@see https://www.php.net/releases/8.2/en.php#deprecate_dynamic_properties

The contrib **facets** module is also flooding Drupal logs with deprecation notices about dynamic property usage.  There's even a [ticket](https://www.drupal.org/project/facets/issues/3360426) for it.  Not sure when that'll be merged.  It doesn't help that the patch is for facets:3.x whereas localgov_directories relies on facets:2.x which is also the stable branch.


stephen-cox (approved):
Tests all pass. I'm happy with this.

Adnan-cds:
Thanks a lot for the approval.  Merging...

PR #319, opened 2023-10-20
Update: localgov_openreferral

Adnan-cds:
Updating localgov_openreferral dependency declaration for Drupal 10 compatibility.


andybroomfield (approved):
Makes sense to me.

PR #325, opened 2023-11-10
[3.x] Facet group OR within AND accross groups

### LINKED ISSUE #193: Use an 'and' query between facet groups and 'or' within them.

andybroomfield:
This is a rebase of #235 and #269 onto the 3.x branch

Fix #193

Adds a new facet plugin to directories  'LocalGov Directories - AND Facet Groups' which when added to the directories facets will perform facet searches using an OR filter within a facet group and an AND filter accross the groups.

This includes the work needed to try to find all possible facets in a group, this is needed as facets only presents facets that are in the result set, but an OR facet in each group needs to check if they would be visible when the other facets apply an AND filter onto them.

Based on work by @snpower, @ekes and @andybroomfield
 
- added 2 new facets plugins for processor and query_type. Together they provide a new pre_query processor on facets to AND facet groups together while still allowing OR within a facet group. This is based on initial code provided by Andy Broomfield
- fixed coding style issues
- coding style tweaks
- Enable And/Or custom facet behaviour by default on Directories Facet.
- Add a test for the facet And groups
- Remove unneeded assertStatusCode
- WIP Filter the facets so that the rest of the facets in group visible
- Add function to fetch sibling facets that would be avalible
- Coding standards.
- Accept no deep clone, clarify what the code is doing.
- Restore by reference for the conditions and fix for Drupal 10
- Adds test for checking that correct facets are visible during selection
- Coding standard fix on comments on the facet groups test
- Reverse the facet group filter removing the passed in facet filter group

### DIFF
--- a/src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
+++ b/src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
@@ -71,7 +71,7 @@ class LocalGovDirectoriesQueryType extends QueryTypePluginBase {
     // Get passed in facets.
     $temp_query = $this->query->getOriginalQuery();
     $temp_query->preExecute();
-    $conditions = &$temp_query->getConditionGroup()->getConditions();
+    $conditions = $temp_query->getConditionGroup()->getConditions();
 
     // Find all the facet condition groups.
     $facet_conditions = [];
@@ -147,7 +147,7 @@ class LocalGovDirectoriesQueryType extends QueryTypePluginBase {
     $filter_query->preExecute();
 
     // Find conditions.
-    $conditions = &$filter_query->getConditionGroup()->getConditions();
+    $conditions = $filter_query->getConditionGroup()->getConditions();
 
     // Store removed conditions so we can reset them.
     $removed_conditions = [];
@@ -172,10 +172,8 @@ class LocalGovDirectoriesQueryType extends QueryTypePluginBase {
     $facets = $filter_query->getResults()->getExtraData('search_api_facets');
     $filter_query->postExecute();
 
-    // Beacuse for reasons unknown, removing the conditions removes it from all
-    // the following queries, even though we are fetching the original query.
-    // We can get around that by readding the conditions back in now that we
-    // have the facets we want.
+    // Without deep cloning we've affected the conditions, reset these for the
+    // query.
     $conditions = array_merge($conditions, $removed_conditions);
 
     // Since we will get every facet from the passed in facet group,
--- a/src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
+++ b/src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
@@ -66,32 +66,43 @@ class LocalGovDirectoriesQueryType extends QueryTypePluginBase {
   public function build() {
     $results = $this->results;
     $query_operator = $this->facet->getQueryOperator();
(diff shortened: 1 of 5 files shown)


andybroomfield:
This should be ready for review now.

finnlewis:
Amazing @andybroomfield thank you!

I will find time to test. 

@snpower sorry it has taken so long since your original pull request. Are you able to take a look at this reworked pull request and confirm it sill works for your use case?

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
```diff
@@ -0,0 +1,201 @@
...
+          $removed_conditions[$cid] = $conditions[$cid];
+          unset($conditions[$cid]);
+        }
+      }
+    }
+
+    // Execute the filter query and get the facets returned.
+    $filter_query->execute();
+    $facets = $filter_query->getResults()->getExtraData('search_api_facets');
+    // dpm($facets);
```
ekes: probably wants to go

ekes (commented):
Has updated changes that were possible from my previous review, so from that perspective all good.

We should probably document how complex the queries are getting, in comparison with using separate configuration facets.

finnlewis:
To test on local: 

 - Install localgov_demo
 - edit localgov_direccories_facets
 - Check the AND Facet Groups option
 - Add some more facets to the organisations
 - The list should then or between facets within a facet and AND between the facets.

finnlewis (approved):
Thanks @andybroomfield - I've tested locally and confirmed this works for me!

How I tested: 

1. Installed localgov_demo
2. Added a new facet type 'Music' and new facet values to the /localgov-drupal-collaborators directory channel.
3. Enabled the option: 

image

4. Confirmed that the facets are now OR'd within the facet group and AND'd between facet groups.

Nice!

src/Plugin/facets/query_type/LocalGovDirectoriesQueryType.php
andybroomfield: Good spot, thats done now.

andybroomfield:
This has now been deployed to https://www.brighton-hove.gov.uk/directories/special-educational-needs-and-disability-send-support

finnlewis:
@ekes sounded happy with this too. 

I think we can merge and release this baby!

PR #328, opened 2023-11-27
fix: added PHP 8.2 to workflow tests

millnut:
I noticed we are not currently checking PHP 8.2 for this module so I've added it to the workflow file.

Added PHP 8.2 compatibility patch for drupal/facets

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,7 @@ jobs:
           - '~10.0'
         php-version:
           - '8.1'
+          - '8.2'
 
     steps:
 
@@ -110,6 +111,7 @@ jobs:
           - '~10.0'
         php-version:
           - '8.1'
+          - '8.2'
 
     steps:
 
@@ -145,6 +147,7 @@ jobs:
           - '~10.0'
         php-version:
           - '8.1'
+          - '8.2'
 
     steps:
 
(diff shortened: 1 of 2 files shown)


millnut:
Looking further at `drupal/facets` it is used across some other localgov modules so it might be preferred to have the patch in the profile

finnlewis:
Error in PHP 8.2: 

```
1) Drupal\Tests\localgov_directories\Functional\FacetsTest::testFacetsGroupFilters
Exception: Deprecated function: Creation of dynamic property Drupal\facets\Result\Result::$transliterateDisplayValue is deprecated
Drupal\facets\Plugin\facets\processor\DisplayValueWidgetOrderProcessor->sortResults()() (Line: 72)
```

@Adnan-cds thinks he's seen this one.

millnut:
@finnlewis yep for some reason the patch doesn't look like it is applying here correctly with the workflow. That error should be fixed by the `drupal/facets` patch; however I think I might revert that and put it in the profile as there are other modules which depend on `drupal/facets` and keep this to a workflow change for PHP 8.2 support only

finnlewis:
Noted. thanks @millnut 

We should probably use the patch, rather than the merge request: 
 
https://www.drupal.org/files/issues/2023-08-16/3336646-function-is-deprecated.patch

Currently it is the same, but a merge request can change.

finnlewis:
Going to try merging this and  test on a release merge request

Issue #340, opened 2024-02-15
Change field formatter default for Enquiries email from text to email


bisd:
This issue is linked to the following issue:
https://github.com/localgovdrupal/localgov_base/issues/496

Concomitant to changing the text of the email address to a mailto in the above issue, I also need to change the field formatter default to email.

bisd:
I have made a new pull request to change the field formatter default against localgov_directories [here](https://github.com/localgovdrupal/localgov_directories/pull/341).

PR #341, opened 2024-02-15
Changed field formatter default for Enquiries email from text to email issue #340

bisd:
This is to complete PR https://github.com/localgovdrupal/localgov_base/pull/523.
It relates to issue https://github.com/localgovdrupal/localgov_directories/issues/340


ekes:
Looks good, but I think we really only want the change to line 74. I suspect the next person coming along will put back `{  }` for `{}` as that's how they export at the moment... or has this changed?

bisd:
Thanks @ekes. VSCode is deleting the spaces within the braces on saving so I opened it in Sublime Text and reverted the spaces. The PR is still failing tests I am not sure why.

ekes (approved):
Failing tests are unrelated to this change.

Works now for setting the email to a link.

ekes:
@bisd nice one thanks.

PR #342, opened 2024-02-22
Fix deprecated conditions in block config

webflo:
The installation of `localgov_microsites` profile fails because of the deprecated (and removed) block visibility plugin. All other blocks in the module use `id: entity_bundle:node` already.


ekes (approved):
Good catch. Thanks.

ekes:
Follow-up https://github.com/localgovdrupal/localgov_directories/issues/346

Issue #346, opened 2024-02-27
Update block conditions (again)


ekes:
Seems a couple of block conditions were overlooked https://github.com/localgovdrupal/localgov_directories/pull/342/files we updated all installs with a [post_update hook](https://github.com/localgovdrupal/localgov_directories/blob/f871834f2e2a60e10d5b2a2ba6a49db889a7d7c6/localgov_directories.post_update.php#L15) but if we've installed more with old config we should update that too! I think just repeating the hook with a new name is safe.

Make sense?

ekes:
Not sure it will have ever been successfully installed since the previous post update hook will have run. But just in case.

andybroomfield:
Do you we know if this is still an issue? Any steps to test?

PR #348, opened 2024-02-27
Check old config doesn't have deprected block visibliity conditions.

ekes:
Just in case #346

### DIFF
--- a/localgov_directories.post_update.php
+++ b/localgov_directories.post_update.php
@@ -26,3 +26,26 @@ function localgov_directories_post_update_replace_node_type_condition() {
     }
   }
 }
+
+/**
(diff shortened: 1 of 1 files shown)


Adnan-cds:
Code changes look good.  Would you mind also fixing this unrelated [coding standard failure](https://github.com/localgovdrupal/localgov_directories/actions/runs/8066134674/job/22033618496?pr=348#step:4:103) from [line 34 of the localgov_directories.post_update.php file](https://github.com/localgovdrupal/localgov_directories/blob/5d6a940bac3b54dadca1e261894946ee3dd7c0e8/localgov_directories.post_update.php#L34).

PR #354, opened 2024-03-08
Update composer.json

AWearring:
Removed patch for `PHP 8.2 compatibility for the 2.0.x branch: https://www.drupal.org/project/facets/issues/3336646` as this has been fixed in 2.0.7 of Drupal Facets Module: https://www.drupal.org/project/facets/releases/2.0.7


millnut (changes requested):
Hi @AWearring it doesn't look like the formatting for this is correct

AWearring:
@millnut Should be sorted now - a case of "more haste and less speed"

ekes:
Should we be sure that it updates then? So contstraining facets to ^2.0.7?

millnut:
> Should we be sure that it updates then? So contstraining facets to ^2.0.7?

Yep 2.0.7 is marked as recommended so I think we can increase the constraint to `^2.0.7` in composer.json

AWearring:
> > Should we be sure that it updates then? So contstraining facets to ^2.0.7?
> 
> Yep 2.0.7 is marked as recommended so I think we can increase the constraint to `^2.0.7` in composer.json

Done

finnlewis (approved):
Looks good to me, thanks @AWearring !

PR #359, opened 2024-03-11
Also add facet configuration on updating the index (enabling)

ekes:
Second part of #355

### DIFF
--- a/config/optional/block.block.localgov_directories_facets.yml
+++ b/config/optional/block.block.localgov_directories_facets.yml
@@ -18,13 +18,14 @@ settings:
   id: 'facet_block:localgov_directories_facets'
   label: 'Directory facets'
   provider: facets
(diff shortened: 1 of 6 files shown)


ekes:
Normally the facet configuration is added when the facets field is added to the index. This doesn't happen if the index isn't enabled. Seems be the case for microsites where it is installed in one go, could be anywhere else though. This hook should check for the field and facet configuration also when the search index is enabled.

finnlewis:
Nice on @ekes  - testing this now.

finnlewis:
Yup, this works better now on a fresh install of LocalGov Microsites. 

Tests pass too... happy to get this merged.

Do we also want to cherry-pick / back-port to 2.x?

finnlewis (approved):
:+1:

PR #373, opened 2024-05-03
fix: dynamic property deprecation

millnut:
Fix for `Creation of dynamic property Drupal\Tests\localgov_directories\Functional\FacetsTest::$channel_node is deprecated`


ekes (approved):
As defined in https://github.com/localgovdrupal/localgov_directories/blob/3d745a23bff58e4cb9a97de04153c926e1f27401/tests/src/Functional/FacetsTest.php#L65

PR #376, opened 2024-05-24
Added PreviewLinkAutopopulate plugin to auto-populate preview links

stephen-cox:
Part of localgovdrupal/localgov#600


finnlewis:
Note depends on https://github.com/localgovdrupal/localgov/pull/713

ekes (approved):
Sensible this only works down, and it works. So approving but one question:

Is this potentially making a button that might crash the site? Unlike step by step and guides can we expect huge numbers of nodes, and at what point is the loadByProperties or the form itself going die? I guess it's not so important.

finnlewis:
Note: it would be great to add a test for performance limitations on this. How many is too many entities?

PR #377, opened 2024-05-24
Use shared workflows for running Github testing

stephen-cox:


finnlewis (approved):
Great! Thanks @stephen-cox

Issue #381, opened 2024-06-17
Use a released version of search_api_location


millnut:
Currently directories is using the dev branch (`1.x-dev`) of `search_api_location` https://github.com/localgovdrupal/localgov_directories/blob/3.x/composer.json#L13

This means any change in an unreleased branch will get included the next time a user runs a `composer install` or upgrade of directories.

Comparing dev to alpha3 tagged release I don't think anything is missing https://git.drupalcode.org/project/search_api_location/-/compare/8.x-1.0-alpha3...8.x-1.x?from_project_id=56597&straight=false

Trying to find some previous context I found https://github.com/localgovdrupal/localgov_directories/pull/207, @ekes do you think it could be switched to the tagged alpha release instead now?

stephen-cox:
We certainly shouldn't be referencing the dev version without pinning to a specific commit, so this seems to be an oversight. If alpha contains everything in dev then I suggest we change the reference.

ekes:
OK that was added by 0325940bfcfb7d7bca4b71709c4efb3b93f2c5bd so it's the proximity search support stuff, that it got lost in. I think what we needed got rolled into a release (which was actually committed a bit soon) but I'll check.

ekes:
It is, but there's been at least two commits since that I think might effect us https://git.drupalcode.org/project/search_api_location/-/commit/132116e6d0bbf53c4c3c111093e6725467ca20de and https://git.drupalcode.org/project/search_api_location/-/commit/16ba62af992ceb17175c0f63b5bc5db8c3560eea

So it might be safer to pin at the latest commit which we know to be working for us cf3e546770099b1f136c020ade79a650d4178ad6

PR #382, opened 2024-06-18
Fix search_api_location to a commit.

ekes:
For required commits see https://github.com/localgovdrupal/localgov_directories/issues/381#issuecomment-2175670250


stephen-cox (approved):
:+1:

PR #390, opened 2024-08-26
checks for $is_syncing

### LINKED ISSUE #389: hook_install should not be making config changes when importing config ($is_syncing == TRUE)

markconroy:
Closes #389


ekes (approved):
All setting form displays that should have been set and exported as cofig.

Issue #370, opened 2024-04-25
Module cannot be reinstalled after uninstall.


rupertj:
If you install, uninstall, then try to re-install this module, the re-install fails:

`
  Configuration objects (core.base_field_override.node.localgov_directory.promote, core.entity_form_display.node.localgov_directory.default, core.entity_view_display.node.localgov_directory.default, core.entity_view_display.node.localgov_directory.search_index, core.entity_view_display.node.localgov_directory.search_result, core.entity_view_display.node.local
  gov_directory.teaser, core.entity_view_mode.node.directory_index, field.field.node.localgov_directory.body, field.field.node.localgov_directory.localgov_directory_channel_types, field.storage.node.localgov_directory_address, field.storage.node.localgov_directory_channel_types, field.storage.node.localgov_directory_channels, field.storage.node.localgov_direc
  tory_email, field.storage.node.localgov_directory_files, field.storage.node.localgov_directory_job_title, field.storage.node.localgov_directory_name, field.storage.node.localgov_directory_notes, field.storage.node.localgov_directory_phone, field.storage.node.localgov_directory_title_sort, field.storage.node.localgov_directory_website, node.type.localgov_dir
  ectory, search_api.index.localgov_directories_index_default, views.view.localgov_directory_channel) provided by localgov_directories already exist in active configuration

Failed to run drush en localgov_directories: exit status 1
`

zackdavies:
PR available for this: https://github.com/localgovdrupal/localgov_directories/pull/392

andybroomfield:
I've added a general issue here https://github.com/localgovdrupal/localgov/issues/762

So news (and publications?) do allow reinstallation, but it seems other content types don't. It also creates an issue whereby if there is existing content they leave inaccesible content so maybe if we do go down this route we need to prevent uninstallation if content exists.

PR #392, opened 2024-08-28
Allow module reinstall

zackdavies:
Adding enforced dependencies to config files to allow reinstallation - see here: https://github.com/localgovdrupal/localgov_directories/issues/370

### DIFF
--- a/config/install/core.base_field_override.node.localgov_directory.promote.yml
+++ b/config/install/core.base_field_override.node.localgov_directory.promote.yml
@@ -1,6 +1,9 @@
 langcode: en
 status: true
 dependencies:
+  enforced:
+    module:
+      - localgov_directories
   config:
     - node.type.localgov_directory
 id: node.localgov_directory.promote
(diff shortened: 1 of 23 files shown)


ekes:
https://github.com/localgovdrupal/localgov/issues/762#issuecomment-2325969467

How much does directories keep on working when you uninstall it, but keep all the config in place? Is it possible someone is wanting to do this? Do we want to support it?

stephen-cox (changes requested):
For this to fully fix the issue described in #370 it also needs to add an enforced dependency to all the submodules in the `modules` directory.

stephen-cox:
Further to my comment above, you are not prevented from uninstalling Directories if directory content exists, leaving phantom content which throws errors if you try to access it. We'll need to address this first https://github.com/localgovdrupal/localgov/issues/762

finnlewis:
Interesting! I think we'll need to test this a bit more.

finnlewis:
Discussing in Merge Tuesday, looking at the suggestion from @millnut on https://github.com/localgovdrupal/localgov/issues/762#issuecomment-2325229657

> We could look at creating a module uninstall validator using the Drupal\Core\Extension\ModuleUninstallValidatorInterface and inform the users that the module cannot be uninstalled as content exists for the news content type. This way it's a user decision to remove all the content before uninstalling the module.


We think we want to follow this, and write a validation to prevent uninstalling until the content has been deleted.

PR #395, opened 2024-09-14
fix: use a tagged release for search_api_location

### LINKED ISSUE #381: Use a released version of search_api_location

millnut:
## What does this change?

Fixes https://github.com/localgovdrupal/localgov_directories/issues/381. We are pinning against a dev version of `search_api_location`, now there is a newer tagged release lets use that https://www.drupal.org/project/search_api_location

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "drupal/pathauto": "^1.6",
         "drupal/search_api": "^1.29",
         "drupal/search_api_autocomplete": "^1.3",
-        "drupal/search_api_location": "1.0-alpha4",
+        "drupal/search_api_location": "1.0.0-alpha4",
         "localgovdrupal/localgov_core": "^2.12",
         "localgovdrupal/localgov_geo": "^2.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "drupal/pathauto": "^1.6",
         "drupal/search_api": "^1.29",
         "drupal/search_api_autocomplete": "^1.3",
-        "drupal/search_api_location": "1.x-dev#cf3e546770099b1f136c020ade79a650d4178ad6",
+        "drupal/search_api_location": "^1.0@alpha",
         "localgovdrupal/localgov_core": "^2.12",
         "localgovdrupal/localgov_geo": "^2.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "drupal/pathauto": "^1.6",
         "drupal/search_api": "^1.29",
         "drupal/search_api_autocomplete": "^1.3",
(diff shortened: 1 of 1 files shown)


stephen-cox:
Rather than referencing this as `"^1.0@alpha"` should we be pinning this to the latest commit reference? Which would make it `1.x-dev#4a13f715d01e74d0858246fd4d691e16491e991c`

millnut:
Nope because we want to use the tagged drupal.org release [8.x-1.0-alpha4](https://www.drupal.org/project/search_api_location/releases/8.x-1.0-alpha4) not the dev branch. 

We was only using the dev branch because of some functionality if I remember correctly, however there is a new tagged release so we no longer need to pin to dev branch

finnlewis:
@ekes thinks we should check the relase includes what we need and than: 

Make a call on whether to track the alpha or pin to alpha4.

And maybe add some tests? 

@Adnan-cds @ekes

millnut:
Hi @Adnan-cds @ekes I wasn't aware of how prone to breaking between releases this module was/is 😄 , so I've pinned it to alpha4

Adnan-cds:
I can confirm that the search_api_location module version 1.0.0-alpha4 works as expected in location search.

composer.json
```diff
@@ -11,7 +11,7 @@
         "drupal/pathauto": "^1.6",
         "drupal/search_api": "^1.29",
         "drupal/search_api_autocomplete": "^1.3",
-        "drupal/search_api_location": "1.x-dev#cf3e546770099b1f136c020ade79a650d4178ad6",
+        "drupal/search_api_location": "1.0-alpha4",
```
Adnan-cds: `1.0-alpha4` works as version constraint but how about `1.0.0-alpha4`?

millnut: Yep that should work, I've updated composer to use this and will see if tests are ok again

Adnan-cds: Thanks.

Adnan-cds (approved):
Thanks :)

ekes (approved):
We'll just have to keep a watch on the next release, and test it before bumping it up.

PR #398, opened 2024-09-27
Feat: Support for the facets_form module

### LINKED ISSUE #326: Support facets form
Support the [facets form ](https://www.drupal.org/project/facets_form) module to allow multiple facets to be seletected without submitting the page.

Currently clicking a facet will cause a page reload (because AJAX support won't work until we upgrade to facets 3).

However it may be useful to allow users to select multiple facets first before filtering the list.
Enabling facets form however breaks the layout of the facets as it presents them differently (the groups arn't present) as it's a different facet block.

Screenshot 2023-11-10 at 3 33 45 pm

Adnan-cds:
While rendering the `facets_form_checkbox` facet widget provided by the facets_form module, groups facet checkboxes according to their LocalGov Directory facet type.  This is similar to how the existing `checkbox` facet widget is rendered for LocalGov Directory facets.

Resolves #326

## What does this change?

Adds support for the [facets_form](https://drupal.org/project/facets_form) module.

## How to test

- Install the facets_form module.
- Edit the localgov_directories_facets facet from `/admin/config/search/facets/localgov_directories_facets/edit`
- Select the `Checkboxes (inside form)` widget near the top of the page and save settings.
- [Optional] Disable the existing facet block from the block layout page at `/admin/structure/block`.
- From the block layout page at `/admin/structure/block`, place the `Facet form: View Directory channel, display Embed` block in a sidebar.
-  Open an existing directory channel page and filter results using the facet checkboxes.

Usual appearance and behaviour expected for LocalGov Directory facets.  Facets should work even without Javascript as we are no longer relying on Javascript to render the checkboxes.  The anchor tags with facet links are also not needed.  This stops crawlers from following these anchor tags.

### DIFF
--- a/css/facets_form.css
+++ b/css/facets_form.css
@@ -1,0 +1,11 @@
+/**
+ * @file
+ * Style rules for facets widgets from the facets_form module.
+ */
+
+/**
+ * Hides the submit buttons on an empty facets form.
+ */
+.facet-empty ~ .form-actions {
+  display: none;
+}
--- a/localgov_directories.libraries.yml
+++ b/localgov_directories.libraries.yml
@@ -8,3 +8,10 @@ localgov_directories_search:
   dependencies:
     - core/drupalSettings
     - core/once
+
+# Hides submit buttons when there are no facet checkboxes to display.
+facets_form_checkboxes_empty:
+  version: 1.x
+  css:
+    state:
+      css/facets_form.css: {}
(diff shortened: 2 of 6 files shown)


Adnan-cds:
I have renamed the Git branch and raised a [new pull request](https://github.com/localgovdrupal/localgov_directories/pull/399).

PR #399, opened 2024-10-02
Feature: Support for the facets_form module

### LINKED ISSUE #326: Support facets form

Adnan-cds:
While rendering the `facets_form_checkbox` facet widget provided by the [facets_form](https://drupal.org/project/facets_form) module, this change groups facet checkboxes according to their LocalGov Directory facet type.  This is similar to how the existing `checkbox` facet widget is rendered for LocalGov Directory facets.

Resolves #326.  Replaces #398 

## What does this change?

Adds support for the [facets_form](https://drupal.org/project/facets_form) module.

## How to test

- Install the facets_form module.
- Edit the localgov_directories_facets facet from `/admin/config/search/facets/localgov_directories_facets/edit`
- Select the `Checkboxes (inside form)` widget near the top of the page and save settings.
- From the block layout page at `/admin/structure/block`, place the `Facet form: View Directory channel, display Embed` block in a sidebar.
- [Optional] Disable the existing facet block from the block layout page at `/admin/structure/block`.
-  Open an existing directory channel page and filter results using the facet checkboxes.

**Note**: The only way to reset all the facet checkboxes is to click the `Reset` button.  Manually unchecking all facet checkboxes followed by a form submission does not work.

Usual appearance and behaviour expected for LocalGov Directory facets.  Facets should work even without Javascript as we are no longer relying on Javascript to render the checkboxes.  The hidden anchor tags with facet links are also not needed.  This stops crawlers from following these anchor tags and then getting stuck in a rabbit hole.


Polynya:
Thanks @Adnan-cds, this works for me.

The patch here resolves the issue with unchecking all the checkboxes.
https://www.drupal.org/node/3255007

Adnan-cds:
Many thanks for testing this Adrian :)

andybroomfield (approved):
Have also tested this in a default localgov drupal install and with BHCC code base.
Can confirm I do see the facets checkboxes, although it is unstyled which I assume is a theme issue? Maybe could provide a facets form template either here or through localgov_base.

Screenshot 2024-10-10 at 7 17 17 PM

Screenshot 2024-10-10 at 7 41 56 PM

Adnan-cds:
> although it is unstyled which I assume is a theme issue? Maybe could provide a facets form template either here or through localgov_base.

Hi Andy, thanks for testing and approving.  There is a new template for the [facet checkboxes](https://github.com/localgovdrupal/localgov_directories/pull/399/files#diff-9810a6ed9391c2da7cb75d2cee112bbe6c4705be2b0e7d8617b26a5f418c050d) but, yes, it is unstyled.  Given that this pull request introduces an optional feature which won't immediately effect any existing site, should I raise a new ticket to style it?  I have to say I am undecided if the Front-end code should go into this repo or localgov_base :)

andybroomfield:
@Adnan-cds I'd do it on a new branch. I think it might need to be in localgov_base as it will need the same classes as the checkboxes.

Adnan-cds:
Okay Andy, I will raise a ticket in localgov_base and leave it to Chris or Mark.

Adnan-cds:
[Related Front-end styling request](https://github.com/localgovdrupal/localgov_base/issues/633).

Issue #412, opened 2025-02-13
Allow Facets 3 upgrade


andybroomfield:
My initial research shows this should work as a straight forward upgrade for directories, keeping the facet blocks in place. The Ajax support requires enabling the facets_exposed_filters module and that has a few more complications but that can be something that finders picks up.

PR #413, opened 2025-02-13
Allow upgrade to facets 3

andybroomfield:
Allows facets to be upgraded to version 3.
This should keep the facet blocks in place from pre upgrade.

## What does this change?

Allows the module upgrade.
That's all that's being enabled here. The extra work with allowing the facets filters module and Ajax support is best left to finders project.

## How to test

Check that facets 3 can be installed

Facets 3 is installable, whilst 2 is still supported.

## Have we considered potential risks?

Still need to throughly check directories works as is with facets 3.

## Images

n/a

## Accessibility

n/a

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "minimum-stability": "dev",
     "require": {
         "drupal/core": "^10.2",
-        "drupal/facets": "^2.0.7",
+        "drupal/facets": "^2.0.7 || ^3.0",
         "drupal/pathauto": "^1.6",
         "drupal/search_api": "^1.29",
         "drupal/search_api_autocomplete": "^1.3",


finnlewis:
This would be good!

tonypaulbarker:
So far, so good. I saw no issues so far for some simple test cases. 

We need to address the failing tests.

We should test mappings of the open referral submodule. 

We should look more into finders (although that specifies its own dependency on Facets 2 so probably not a blocker).

tonypaulbarker:
We also have some views / search configuration to explore:

New in Facets 3: For new sites it is recommended to use the "Facets Exposed Filters" submodule for Facets on Views. This module uses native Views filters instead, which has many advantages.
[Click here](https://project.pages.drupalcode.org/facets/exposed_filters) for documentation on how to use this new workflow.
Adding Facets as a block [is still supported](https://project.pages.drupalcode.org/facets/facet_blocks_support) for backwards compatibility.

andybroomfield:
@tonypaulbarker the purpose here is just to allow sites to perform the upgrade, but not change the default config here from the old style facets block. It just allows site that do want to make this change (to use Ajax filters) to allow them to do so. I'd suggest this is a documentation task for directories 3.x. For 4.x which uses finders. I'd suggest that does default to using facets 3 and the views facets filters.

tonypaulbarker:
That makes sense @andybroomfield - unless something else crops up let's just make sure then that Open Referral mappings will still work.

finnlewis:
See #412 

Note: we don't know of anyone using the open referral sub module in the wild. 

So to test, we might just need to confirm that the module installs and works with facets 3.

tonypaulbarker:
Looks good to me. But I did find a bug also present with facets 2 where using the back button to get back to the news listing from an article prevents selecting facets. Also reported on https://github.com/localgovdrupal/localgov_news/pull/140

andybroomfield:
Please see https://github.com/localgovdrupal/localgov_news/issues/142#issuecomment-2978369142 before merging. If we want to apply the patch mentioned there we need to pick a version of facets (or highlight the two patches in release notes).

PR #434, opened 2025-06-26
Add config for simple sitemap

### LINKED ISSUE #40: Add Simple Sitemap settings to directory nodes

markconroy:
Closes #40 

## What does this change?
Adds default config for simple sitemap - same approach as LocalGov News.

## How to test

Install the directories modules, make sure directories content types have default sitemap settings.


stephen-cox (changes requested):
There are a few static analysis failures - just looks like missing function return types - but these will need to be fixed before this can be merged.

stephen-cox (approved):
Looking good now

PR #436, opened 2025-06-27
Remove leaflet patch

### LINKED ISSUE #435: Leaflet patch no longer applies
Leaflet module has been updated to include the patch added in https://github.com/localgovdrupal/localgov_directories/pull/336

> Cannot apply patch Fix leaflet schema for missing properties on leaflet_popup: https://www.drupal.org/project/leaflet/issues/3473537 (https://git.drupalcode.org/issue/leaflet-3473537/-/commit/1f8d30a70ff7ac07cf7e8537aad65559f83905f2.patch)!

[https://www.drupal.org/project/leaflet/releases/10.2.50](https://www.drupal.org/project/leaflet/releases/10.2.50)
[https://www.drupal.org/project/leaflet/releases/10.3.4](https://www.drupal.org/project/leaflet/releases/10.3.4)

AWearring:
Updated composer.json to remove patch that is failing to apply as it has been included - fixes https://github.com/localgovdrupal/localgov_directories/issues/435

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,9 +30,6 @@
         "patches": {
             "drupal/facets": {
                 "Don't render facet block if backend isn't available: https://www.drupal.org/project/facets/issues/3311856": "https://git.drupalcode.org/issue/facets-3311856/-/commit/765d5ef4228906c7f201e116763f3018a7867c96.patch"
-            },
-            "drupal/leaflet": {
-                "Fix leaflet schema for missing properties on leaflet_popup: https://www.drupal.org/project/leaflet/issues/3473537": "https://git.drupalcode.org/issue/leaflet-3473537/-/commit/1f8d30a70ff7ac07cf7e8537aad65559f83905f2.patch"
             }
         }
     }
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     "require": {
         "drupal/core": "^10.2 || ^11.0",
         "drupal/facets": "^2.0.7 || ^3.0",
-        "drupal/leaflet": "^10.2",
+        "drupal/leaflet": "^10.3",
         "drupal/pathauto": "^1.6",
         "drupal/search_api": "^1.29",
         "drupal/search_api_autocomplete": "^1.3",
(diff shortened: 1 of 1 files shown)


tonypaulbarker:
Let's look at bumping the dependencies too so we get the correct version with the patch included when we remove the patch here. At the moment geo_entity requires ^10.0 

@stephen-cox do you have an opinion on adding a dependency here or making a new release of geo_entity and requiring that?

stephen-cox:
> Let's look at bumping the dependencies too so we get the correct version with the patch included when we remove the patch here. At the moment geo_entity requires ^10.0
> 
> @stephen-cox do you have an opinion on adding a dependency here or making a new release of geo_entity and requiring that?

There's actually a dependency in the Directories Location module on the drupal/leaflet module, which is missing from the composer.json file. I think we should add the dependency on ^10.2 here.

There's a bunch of schema failures which seem to result from this patch being removed - looking into those now.

stephen-cox:
The test failures are coming from the schema in the Geo Entity module. I have create an issue with a fix for this here: https://www.drupal.org/project/geo_entity/issues/3532779

I have also added a dependency on leaflet 10.2 to this module as I believe the dependency should have been stated here in the first place and as this is where we require the patch this is where the version requirement should live.

@AWearring how critical is this issue? Given the changes it would be good to get another set of eyes on this, which means it won't be merged and released before Tuesday.

AWearring:
@stephen-cox It's putting our deployments on hold while the patch fails to apply

PS I'm "awearring" on Drupal not "Ali_W", I seem to have 2 accounts

tonypaulbarker:
I do think we need to review and test this one.
 "drupal/leaflet": "^10.2", is going to allow versions that don't have this patch.
 How about "drupal/leaflet": "^10.2.50,<10.3.0|^10.3.4" ?

millnut:
Given drupal/leaflet `10.3` supports Drupal 10 and 11 I think we should just lock to a minimum of 10.3 and not worry about extra constraints or additional patch handling

stephen-cox:
> Given drupal/leaflet `10.3` supports Drupal 10 and 11 I think we should just lock to a minimum of 10.3 and not worry about extra constraints or additional patch handling

Sounds sensible - I have made the change

finnlewis:
Nice one @stephen-cox - agreed, I was just about to do the same but saw your commit.

PR #440, opened 2025-07-22
Better exposed filter support

andybroomfield:
WIP for #439

Adds initial support for using directory facets with better exposed filters.

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### DIFF
--- a/localgov_directories.module
+++ b/localgov_directories.module
@@ -58,6 +58,10 @@ function localgov_directories_theme() {
     'checkboxes__localgov_directories_facets' => [
       'base hook' => 'checkboxes',
     ],
+    'bef_checkboxes_directory_facets' => [
+      'render element' => 'element',
+      'base hook' => 'bef_checkboxes',
+    ],
   ];
 }
 
@@ -313,4 +317,26 @@ function localgov_directories_set_default_permissions(): void {
   // Default grant permissions to view directory facets.
   user_role_grant_permissions(RoleInterface::ANONYMOUS_ID, ['view directory facets']);
   user_role_grant_permissions(RoleInterface::AUTHENTICATED_ID, ['view directory facets']);
+}
+
+/**
+ * Implements hook_template_preprocess_bef_checkboxes_directory_facets().
+ */
+function template_preprocess_bef_checkboxes_directory_facets(array &$variables): void {
+  $element = $variables['element'];
+  $facet_groups = $element['#localgov_directory_facet_groups'];
+  $facet_group_elements = [];
+
+  // Split up the facet checkboxes into groups based on the facet group.
+  foreach ($facet_groups as $key => $options) {
+    $facet_group_elements[$key] = [
+      '#type' => 'fieldset',
+      '#title' => $key,
+    ];
+    $options = array_keys($options);
+    $facet_group_elements[$key]['children'] = array_map(function ($item) use ($element) {
+      return $element[$item];
+    }, $options);
+  }
+  $variables['localgov_directory_facet_groups'] = $facet_group_elements;
 }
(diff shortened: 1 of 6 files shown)


andybroomfield:
Note: Naming the facets as `localgov_directories_facets` causes the checkboxes to be themed using the facets form template `modules/contrib/localgov_directories/templates/checkboxes--localgov-directories-facets.html.twig` instead.

andybroomfield:
Here is a patch if you want to try better exposed filters with localgov directories
https://gist.github.com/andybroomfield/0b17ba2a48769f6aa031ca9c12652088

src/Plugin/better_exposed_filters/filter/DirectoryFacetsCheckboxes.php
```diff
@@ -0,0 +1,83 @@
...
+   */
+  public function __construct(array $configuration, string $plugin_id, $plugin_definition, ClassResolverInterface $class_resolver) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->classResolver = $class_resolver;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition): DirectoryFacetsCheckboxes {
```
andybroomfield: This needs to be changed to `:static`
Screenshot 2025-11-24 at 12 11 45

andybroomfield:
This has been fixed up to work with the new version of Better exposed filters. 
@localgovdrupal/maintainers @localgovdrupal/core-team is there anything you can think of that we would need to do to make this mergeable. I could look at tests but don't know if that will conflict with facet forms, as it needs facets 3.

localgov_directories.module
```diff
@@ -314,3 +318,26 @@ function localgov_directories_set_default_permissions(): void {
...
+ */
+function template_preprocess_bef_checkboxes_directory_facets(array &$variables): void {
+  $element = $variables['element'];
+  $facet_groups = $element['#localgov_directory_facet_groups'];
+  $facet_group_elements = [];
+
+  // Split up the facet checkboxes into groups based on the facet group.
+  foreach ($facet_groups as $key => $values) {
+    $facet_group_elements[$key] = [
+      '#type' => 'fieldset',
```
rupertj, line 333: How about using details/summary instead of a fieldset here? They are apparently easier to style, and you get the ability to collapse them for free. Drupal's been moving fieldsets to them elsewhere too: https://www.drupal.org/node/1852020

rupertj (commented):
I'm trying this out - it's working well so far. ~The only snag has been that H&F uses collapsible facets on their directories (EG https://www.lbhf.gov.uk/living-independently/living-independently-service-directory) so I'd need to be able to recreate that.~

localgov_directories.module
andybroomfield: @rupertj I think for now I'm trying to mirror the existing checkboxes set up which uses fieldset. Note that the out the box LGD [does not have collapsible filters](https://demo.localgovdrupal.org/localgov-drupal-collaborators). 

For the finders implementation we could change this if exposed filters and facets 3 becomes the norm.
I think it is possible to change them to details with a themes implementation of hook_preprocess_bef_checkboxes_directory_facets?

rupertj: That makes sense. I'll try it in our own theme then.

Issue #417, opened 2025-02-25
Fix ESLint errors


finnlewis:
See 

1. https://github.com/localgovdrupal/localgov_directories/actions/runs/13395587464/job/37783475399?pr=408
2. https://github.com/localgovdrupal/localgov_directories/actions/runs/13395587464/job/37783474393?pr=408

millnut:
https://github.com/localgovdrupal/localgov_directories/pull/418 should fix the yaml errors

finnlewis:
Still got https://github.com/localgovdrupal/localgov_directories/actions/runs/15345607521/job/43181133432?pr=428#step:4:8

PR #442, opened 2025-07-25
Move populating title sort field later in the index process.

ekes:
See https://github.com/localgovdrupal/localgov_microsites/issues/524

## @todo

Needs update hook testing to enable the processor on existing indexes.

## What does this change?

Moves populating the title sort field into a processor, rather than in item hook. The hook is called earlier and was causing some field item that caused issues with microsites indexing.

## How to test

There is an automated Kernel test for this field being populated. It still passes (locally).

### DIFF
--- a/config/install/search_api.index.localgov_directories_index_default.yml
+++ b/config/install/search_api.index.localgov_directories_index_default.yml
@@ -2,9 +2,6 @@ langcode: en
 status: false
 dependencies:
   config:
-    - field.storage.node.localgov_directory_channels
-    - field.storage.node.localgov_directory_facets_select
-    - field.storage.node.localgov_directory_title_sort
     - core.entity_view_mode.node.directory_index
   module:
     - node
@@ -20,7 +17,7 @@ field_settings:
     type: text
     configuration:
       roles:
-        anonymous: anonymous
+        - anonymous
       view_mode:
         'entity:node':
           localgov_directories_page: directory_index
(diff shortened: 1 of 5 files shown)


ekes:
Update hook added, but not tested.

nccchris (commented):
The search now finds the examples I posted and newly created content: i.e. it's working as expected.

finnlewis:
@nccchris I'm still not getting search results back for content in the text paragraph within the layout paragraph.

Steps I'm taking: 

1. Fresh install, with this branch
2. Create 2 microsites
3. Create page content with layout paragraphs content including "apple banana pear"
4. Check everything is indexed with `ddev drush sapi-i`
5. https://localgov-micro-1.ddev.site/search?s=apple
6. No results

What am I missing?

ekes:
> What am I missing?

As you only mention this branch. Have you also checked out the microsites group branch too? See referenced PR and/or OP that links to both.

finnlewis:
Ah - thanks @ekes , missed that!

Yes, it works with https://github.com/localgovdrupal/localgov_microsites_group/pull/545/files 

:)

finnlewis:
Could do with fixing the phpcs and phpstan checks so the tests run.

finnlewis (approved):

finnlewis:
https://github.com/localgovdrupal/localgov_directories/issues/417 for the ESLint issues

PR #447, opened 2025-09-01
fix/3.x 446 facets view update remove canioncal route

### LINKED ISSUE #446: Directory facets label not viewable to anon users means facets disappear post security update
This security update https://www.drupal.org/sa-contrib-2025-099 added access checking to displaying of facets when set as exposed filters (and possibly the traditional blocks).
However this caused an issue whereby facets disappeared from directories when viewed without being logged in.

For now we have given anon users view permission, although that actually gives view permission to the entity itself. Really we should use the view_label permission instead.

andybroomfield:
Fix #446 

- Remove the canonical route for directory facets
- Allow anon and auth users to view directory facets

## What does this change?

Allow anonymous and authenticated users to view directory facets again, after the security update to facets 3.0.1 / 2.0.10

## How to test

- Run the update and view a directory channel as an anonymous user.
- Try to edit and add directory facets and check functionality is still present.

Directory facets are viewable again.

## Have we considered potential risks?

Some sites might be using the directory facets canonical route.

## Images

Before:
Screenshot 2025-09-01 at 12 12 01 pm

After:
Screenshot 2025-09-01 at 12 12 37 pm

## Accessibility

n/a

### DIFF
--- a/localgov_directories.install
+++ b/localgov_directories.install
@@ -38,6 +38,9 @@ function localgov_directories_install($is_syncing) {
       'priority' => '0.5',
     ]);
   }
+
+  // Allow anonymous and authenticated users to view directory facets.
+  localgov_directories_set_default_permissions();
 }
 
 /**
@@ -314,4 +317,15 @@ function localgov_directories_update_8008() {
   ) {
     \Drupal::keyValue('system.schema')->set('localgov_directories_db', 8000);
   }
+}
+
+/**
+ * Allow anonymous and authenticated users to view directory facets.
+ */
+function localgov_directories_update_8009(): void {
+  localgov_directories_set_default_permissions();
+
+  // Clear caches so removal of directory facets entity canonical route is
+  // picked up.
+  drupal_flush_all_caches();
 }
--- a/localgov_directories.links.task.yml
+++ b/localgov_directories.links.task.yml
@@ -1,15 +1,11 @@
-entity.localgov_directories_facets.view:
-  title: View
-  route_name: entity.localgov_directories_facets.canonical
-  base_route: entity.localgov_directories_facets.canonical
 entity.localgov_directories_facets.edit_form:
   title: Edit
   route_name: entity.localgov_directories_facets.edit_form
-  base_route: entity.localgov_directories_facets.canonical
+  base_route: entity.localgov_directories_facets.edit_form
 entity.localgov_directories_facets.delete_form:
   title: Delete
   route_name: entity.localgov_directories_facets.delete_form
-  base_route: entity.localgov_directories_facets.canonical
+  base_route: entity.localgov_directories_facets.edit_form
   weight: 10
 entity.localgov_directories_facets.collection:
   title: Directory Facets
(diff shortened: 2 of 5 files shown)


tonypaulbarker:
Have been attempting to test.  Both with and without the changes here I observe on a clean copy locally with demo content as user 1 empty directories channels with only the search box visible. Content appears to have installed and reference the channels with no errors apparent.

andybroomfield:
@tonypaulbarker I found that on fresh installs the directory search index needs to be run via search api.

tonypaulbarker:
@andybroomfield of course, that’s probably it. Thanks!

tonypaulbarker:
@andybroomfield 

The permission change appears to resolve the underlying issue tested on a fresh demo install.

There doesn't seem to be a material change for anonymous users. They appear to get a 'page not found' response instead of 'access denied' when visiting, say /admin/content/directories/facets/5. I don't reckon that makes a whole lot of difference to performance in terms of logging and crawlers.

There is a problem for editors, however and the answer to "Would anyone actually visit this page?" looks to be "yes":
Whenever adding a new facets or saving a facet after updating, say /admin/content/directories/facets/3/edit
when saving editors will see an error:

The website encountered an unexpected error. Try again later.

Drupal\Core\Entity\Exception\UndefinedLinkTemplateException: No link template 'canonical' found for the 'localgov_directories_facets' entity type in Drupal\Core\Entity\EntityBase->toUrl() (line 211 of core/lib/Drupal/Core/Entity/EntityBase.php).

Instead of the proposed changes could a role check for these routes work instead? I think we only care about accessing the facet for anonymous users?

tonypaulbarker (changes requested):
We need to address the above error when adding or updating facets

andybroomfield:
@tonypaulbarker I found the culprit to this was that we generate a logger entry when a facet is saved and add a link to view the entity, which is no longer there. I've updated so we link to edit it now.

Screenshot 2025-09-02 at 11 53 25 am

That allows facets to be edited and saved again. 
It does raise a question which I noted in the commit message, any call to 
```php
$entity->toLink()->toRenderable()
```
will fail with a directory facet with this change, because no canonical route is defined, and this is the [default returned](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityInterface.php/function/EntityInterface%3A%3AtoLink/11.x).

It looks like it's quite common to remove [unwanted](https://git.drupalcode.org/project/entity_browser/-/commit/893ea180f44e6590cf9b85d2393404ab7c66b920) [canonical](https://www.drupal.org/project/address_formatter/issues/3469118) [routes](https://www.drupal.org/project/entity_share/issues/3419256) from entities if there not meant to have user facing pages.

Sounds like this needs a bit of discussion to go forward. Mean time we should recommend sites enable the view directory facets permission, even at the cost of the page being visible as otherwise they cannot upgrade to facets 2.0.10 or facets 3.0.1 and will have security warnings.

We can then decide if it is this solution or the solution I implemented for Alert banner which we forward with.

andybroomfield:
If we go ahead with this we should add a test?

finnlewis:
@tonypaulbarker testing again... once tested, if all OK, we're happy to go ahead and get this out asap so people don't lose their facets.

tonypaulbarker (approved):
The add and edit issue looks resolved. Thanks @andybroomfield

PR #451, opened 2025-10-29
Fixes prefixes (subdirectory, language) being included in parent path pattern for other instances.

ekes:
See https://github.com/localgovdrupal/localgov_services/issues/335#issuecomment-3461396195 and https://github.com/localgovdrupal/localgov_services/issues/335


andybroomfield:
I've assigned myself as a reviewer as I think we missed this one.

andybroomfield (approved):
I've tested this in an install in a subdirectory and verified that the url is being updated correct for the channel url when this added to the parent and using auto aliases.

